PR #8, opened 2020-01-13
Several fixes after a flight

olafurpg:


olafurpg:
Also took the first steps towards setting up a docusaurus site https://github.com/scalameta/munit/compare/master...olafurpg:docs?expand=1 but ran out of battery and that one needs a bit more work

PR #11, opened 2020-01-15
sbt 1.3.7

sullis:


olafurpg:
Thank you for this contribution @sullis. I would prefer to enable Scala Steward for this repository when we want to get automatic PRs to update sbt and other dependencies. Currently, we're just exploring some concepts around testing so it's not critical to be on the latest sbt.

PR #12, opened 2020-01-15
Cross-compile to dotty

gabro:
Not fully done yet but it's been a fun exercise


build.sbt
```diff
@@ -17,27 +18,21 @@ inThisBuild(
...
+    crossScalaVersions := List(scala213, scala212, scala211, dotty),
     fork := true,
     testFrameworks := List(
       new TestFramework("munit.Framework")
     ),
     resolvers += Resolver.sonatypeRepo("public"),
-    useSuperShell := false
+    useSuperShell := false,
+    scalacOptions ++= List(
+      "-language:implicitConversions"
```
gabro: Dotty complains a lot without this

build.sbt
```diff
@@ -17,27 +18,21 @@ inThisBuild(
...
 
-lazy val fansiVersion = Def.setting[String] {
-  if (scalaVersion.value.startsWith("2.11")) "0.2.6"
-  else "0.2.7"
-}
-lazy val pprintVersion = Def.setting[String] {
-  if (scalaVersion.value.startsWith("2.11")) "0.5.4"
-  else "0.5.7"
-}
-
```
gabro: These were unused

build.sbt
```diff
@@ -68,9 +64,14 @@ lazy val munit = project
...
       "com.geirsson" % "junit-interface" % "0.11.6",
-      "com.googlecode.java-diff-utils" % "diffutils" % "1.3.0",
-      "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided
-    )
+      "com.googlecode.java-diff-utils" % "diffutils" % "1.3.0"
+    ) ++ {
+      scalaBinaryVersion.value match {
+        case "0.21" => Nil
+        case _ =>
+          List("org.scala-lang" % "scala-reflect" % scalaVersion.value)
```
gabro: I couldn't get the tests to pass with `Provided`, this is probably something bad in my setup

munit/src/main/java/munit/Ignore.java
```diff
@@ -4,5 +4,5 @@
 import java.lang.annotation.Retention;
 
 @Retention(RetentionPolicy.RUNTIME)
-@interface Ignore {
+@interface IgnoreSuite {
```
gabro: Dotty complains because there's already a `munit.Ignore` object (defined in Scala).

I actually think the error makes sense, so I took the liberty of renaming this

munit/src/main/scala-0.21/munit/internal/Compat.scala
```diff
@@ -0,0 +1,25 @@
+package munit.internal
+
+import munit.Location
+import scala.quoted._
+import scala.reflect.Selectable.reflectiveSelectable
```
gabro: took me a while to find, but this is the `Any => Selectable` implicit conversion, which enables the cast to a structural type in `collectionClassName` (in Dotty structural types must extend `Selectable`)

munit/src/main/scala-0.21/munit/internal/Compat.scala
```diff
@@ -0,0 +1,25 @@
...
+  trait LocationMacro {
+    inline implicit def generateLocation: Location = ${ locationImpl() }
+  }
+
+  def locationImpl()(given qctx: QuoteContext): Expr[Location] = {
+    import qctx.tasty.{_, given}
+    val path = rootPosition.sourceFile.jpath.toString
+    val startLine = rootPosition.startLine
+    '{ new Location(${Expr(path)}, ${Expr(startLine)}) }
+  }
```
gabro: So this is quite straightforward, but it seems that the expansion site is slightly different than Scala 2 macros, so some tests fail because the `startLine` is off. I've tried playing around with by-1 fixes, but it's not that simple.

munit/src/main/scala/munit/Assertions.scala
```diff
@@ -1,9 +1,14 @@
 package munit
 
-import org.junit.AssumptionViolatedException
 import scala.reflect.ClassTag
 import scala.util.control.NonFatal
 
+// FIXME(gabro): Constructing a `org.junit.AssumptionViolatedException` causes Dotty to crash
+// so we use our own Exception class to work around it.
+// See https://github.com/lampepfl/dotty/issues/7990
```
gabro: I got real lucky here :P

munit/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -5,18 +5,17 @@ import org.junit.runner.notification.RunNotifier
...
 import org.junit.runner.notification.Failure
 import scala.util.control.NonFatal
-import org.junit.AssumptionViolatedException
 import org.junit.runner.manipulation.Filterable
 import org.junit.runner.manipulation.Filter
 import org.junit.runner.Runner
 import scala.collection.mutable
 
-class MUnitRunner(cls: Class[_ <: Suite]) extends Runner with Filterable {
+class MUnitRunner(val cls: Class[_ <: Suite]) extends Runner with Filterable {
```
gabro: `cls` is references by public members, so it needs to be public. Dotty is quite strict about this and I think it makes sense.

I've also considered to make everything else private, but I got lazy.

munit/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -5,18 +5,17 @@ import org.junit.runner.notification.RunNotifier
...
   require(
     hasEligibleConstructor(),
     s"Class '${cls.getCanonicalName()}' is missing a public empty argument constructor"
   )
-  lazy val suite = cls.newInstance()
+  val suite = cls.newInstance()
```
gabro: speaking of `lazy`, Dotty complains because `suite` is referenced in type position (e.g. `suite.Test`) so it can't be a `lazy` val.

```
[error] -- Error: /Users/gabro/buildo/munit/munit/src/main/scala/munit/MUnitRunner.scala:127:12
[error] 127 |      test: suite.Test
[error]     |            ^^^^^
[error]     | (MUnitRunner.this.suite : MUnitRunner.this.cls.T) is not a legal path
[error]     | since it refers to nonfinal lazy value suite
```

Making it `final` doesn't help:

```
[error] -- Error: /Users/gabro/buildo/munit/munit/src/main/scala/munit/MUnitRunner.scala:127:12
[error] 127 |      test: suite.Test
[error]     |            ^^^^^
[error]     |(MUnitRunner.this.suite : MUnitRunner.this.cls.T) is not a legal path
[error]     |since its underlying type (MUnitRunner.this.suite : MUnitRunner.this.cls.T) is not a concrete type
```

It intuitively make sense to me, and removing the `lazy` was the easiest thing to make it compile.

munit/src/main/java/munit/Ignore.java
olafurpg: Yeah, I can live with this. We should probably turn `val Ignore = new Tag("Ignore")` into `object Ignore extends Tag("Ignore")` so that it can be auto-imported in IDEs

munit/src/main/scala/munit/MUnitRunner.scala
olafurpg: I would prefer to keep the `lazy` modifier and instead case to `Any` or `val suite: Suite { type TestValue = Any }` or something like that.

olafurpg:
`FrameworkSuite` is currently only enabled for 2.13 because of the new `productElementName` and the differences in dotty macro `Position.startLine`. We can fix those down the road and enable `FrameworkSuite` for more Scala versions. This is already a great start to get the cross-build setup. Thank you @gabro ! 🙏

PR #30, opened 2020-01-28
dbuild friendly Scala version checking

SethTisue:
context: adding munit to the community build

I don't actually understand why e.g. `scalaBinaryVersion == "2.11"`
isn't working in this context, but it doesn't seem worth digging into

you might prefer to redo the details of this in your own style, idk


SethTisue:
gah, wait, this works for the 2.11 and 2.12 community build, but not yet in the 2.13 build. investigating...

SethTisue:
oh, actually, I think it's dbuild's fault, something about it not finding `shared` in cross-built projects, idk why, but I can resolve it in our config, this PR is fine

PR #44, opened 2020-02-07
Upgrade Dotty version  to 0.22.0-RC1

gabro:
The upgrade went smooth, and I was able to remove the workaround for https://github.com/lampepfl/dotty/issues/7990, which is now fixed. 

By the way, I suppose we don't want to cross publish for different dotty versions, so I've just dropped 0.21.0-RC1.


gabro:
Tests are failing on Scala Native with this error:

```
[error] /home/runner/work/munit/munit/munit/shared/src/main/scala/munit/Assertions.scala:10:8: object AssumptionViolatedException is not a member of package org.junit
[error] import org.junit.AssumptionViolatedException
[error]        ^
[error] /home/runner/work/munit/munit/munit/shared/src/main/scala/munit/Assertions.scala:41:19: not found: type AssumptionViolatedException
[error]         throw new AssumptionViolatedException(munitPrint(clue))
[error] 
```

@olafurpg any idea why?

gabro:
Ah, looking at the build it's obvious: junit is not a dependency on native. Was that intentional? I know nothing about scala native, and it shows...

gabro:
Ok, I turned my brain up and pushed a fix. It hopefully makes sense

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -37,7 +38,7 @@ trait Assertions {
   )(implicit loc: Location): Unit = {
     StackTraces.dropInside {
       if (!cond) {
-        throw new DottyBugAssumptionViolatedException(munitPrint(clue))
```
olafurpg: Should we delete `DottyBugAssumptionViolatedException`?

olafurpg: Nevermind, the diff was collapsed. I see it's deleted now

gabro: didn't I do it?

PR #49, opened 2020-02-12
Added message check to intercept assertion

gzoller:
Added a check to existing intercept assertion to (optionally) validate that the expected exception's message matches an expected value.  If no message is given, intercept() works as it does today.

Added documentation for intercept().

### DIFF
--- a/munit/shared/src/main/scala/munit/Assertions.scala
+++ b/munit/shared/src/main/scala/munit/Assertions.scala
@@ -120,7 +120,7 @@ trait Assertions {
           else {
             val obtained = e.getClass().getName()
             throw new FailException(
-              s"intercept failed, exception '$obtained' had message '${e.getMessage}', which was different from expected message '$msg'",
+              s"intercept failed, exception '$obtained' had message '${e.getMessage}', which was different from expected message '${msg.get}'",
               cause = e,
               isStackTracesEnabled = false,
               location = loc
--- a/munit/shared/src/main/scala/munit/Assertions.scala
+++ b/munit/shared/src/main/scala/munit/Assertions.scala
@@ -92,17 +92,17 @@ trait Assertions {
   def intercept[T <: Throwable](
       body: => Any
   )(implicit T: ClassTag[T], loc: Location): T = {
-    _intercept(None, body)
+    runIntercept(None, body)
   }
 
-  def interceptMessage[T <: Throwable](msg: String)(
+  def interceptMessage[T <: Throwable](expectedExceptionMessage: String)(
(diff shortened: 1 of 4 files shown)


munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -90,7 +90,8 @@ trait Assertions {
   }
 
   def intercept[T <: Throwable](
```
olafurpg: Thank you for this contribution! Is this the same signature as in scalatest? I feel like it would be better to have an interceptMessage assertion and leave intercept unchanged. What do you think?

gzoller: No worries!  No, scalatest has a different API, I think.  Most frameworks have some way to validate an Exceptions message, as this is very useful.  I added interceptMessage() as suggested because the API is cleaner.  Now you can say:
```scala
interceptMessage[MyException]("boom"){ 
  // scary code here
}
```

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -90,8 +90,20 @@ trait Assertions {
...
-      body: => Any,
-      msg: String = ""
+      body: => Any
+  )(implicit T: ClassTag[T], loc: Location): T = {
+    _intercept(None,body)
+  }
+
+  def interceptMessage[T <: Throwable](msg: String)(
```
olafurpg: How about?
```suggestion
  def interceptMessage[T <: Throwable](expectedExceptionMessage: String)(
```

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -90,8 +90,20 @@ trait Assertions {
...
+      body: => Any
+  )(implicit T: ClassTag[T], loc: Location): T = {
+    _intercept(Some(msg),body)
+  }
+
+  private def _intercept[T <: Throwable](
```
olafurpg: It's not idiomatic to use `_name` in Scala
```suggestion
  private def runIntercept[T <: Throwable](
         body: => Any,
         expectedExceptionMessage: Option[String]
```

gzoller: done

munit/shared/src/main/scala/munit/Assertions.scala
gzoller: done

gzoller:
Hi Ólafur, are we good to merge this change?

PR #59, opened 2020-02-17
 Add failSuite() helper to abort the entire test suite

### LINKED ISSUE #37: Add failSuite() helper to abort the entire test suite
Currently, you can use `fail()` to abort a single test case with a custom message. However, the next test case will continue to run unchanged.

It would be nice if there was a `failSuite()` helper that would abort subsequent test cases as well.

Context: https://contributors.scala-lang.org/t/announcing-munit-a-new-scala-testing-library/4007/5?u=olafurpg

daron666:
Fixes #37. Previously, it was possible to abort an individual test case based on a dynamic condition but it was not possible to abort the entire test suite. Now, users can call `failSuite("abort test suite!")` to prevent subsequent test cases from starting.

### DIFF
--- a/munit/shared/src/main/scala/munit/MUnitRunner.scala
+++ b/munit/shared/src/main/scala/munit/MUnitRunner.scala
@@ -31,6 +31,7 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
   private val suiteDescription = Description.createSuiteDescription(cls)
   private implicit val ec: ExecutionContext = suite.munitExecutionContext
   @volatile private var filter: Filter = Filter.ALL
+  @volatile private var suiteAborted: Boolean = false
   val descriptions: mutable.Map[suite.Test, Description] =
     mutable.Map.empty[suite.Test, Description]
   val testNames: mutable.Set[String] = mutable.Set.empty[String]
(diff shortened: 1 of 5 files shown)


munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -200,6 +201,16 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
     if (!filter.shouldRun(description)) {
       return Future.successful(false)
     }
+    if (suiteAborted) {
+      notifier.fireTestAssumptionFailed(
+        new Failure(
+          description,
+          new IllegalStateException("Suite has been aborted.")
```
olafurpg: Can we use `FailSuiteException`?

olafurpg: Let's remove the dot
```suggestion
          new IllegalStateException("Suite has been aborted")
```

olafurpg:
PS. I took the liberty to update the PR description with more details. For future PRs, it's good to follow the "previously, ... Now, ... new behavior" structure to help readers understand what changed and why.

munit/shared/src/main/scala/munit/MUnitRunner.scala
daron666: Done!

daron666:
@olafurpg omg, thanks for editing this PR. Next time will do it by myself. 👍

PR #61, opened 2020-02-23
Merge `munitTestValue` with `munitRunTest`

kubukoz:
As suggested [on gitter](https://gitter.im/scalameta/munit?at=5e4ecce24880f07ed1e9fb6a).


olafurpg:
Thank you for this contribution! Sorry it took this long to respond, I've been taking time off.

I opened https://github.com/scalameta/munit/pull/64 as an alternative proposal that revamps a lot of the design around how to customize how tests run. I would be curious to know what you think

olafurpg:
Both `munitTestValue` and `munitRunTest` have been removed in favor of test, suite and value transforms in #64

Issue #67, opened 2020-03-18
Update build to support Dotty 0.23.0-RC1


gzoller:
This is a tracking issue.  Dotty 0.23.0-RC1 has been tagged and is due to drop imminently.  There are some minor code changes required to support the move from 0.22 to 0.23.  I have a PR ready to submit as soon as it drops.  This issue is to track and raise awareness for this change, and can be closed if/when the PR is merged.

gzoller:
PR submitted and CI tests passed.

gzoller:
I also just noticed that there are not Dotty builds released into maven central.  Is that true?  It'd be super-nice if these were published.  Right now munit is really the only workable test framework for Dotty, so it'd be very handy for those of us developing Dotty-based projects to have munit as a downloadable dependency.

gzoller:
Actually--you're right.  I re-tried my build with "using" back in and it worked.   Pushing update in a few minutes.

olafurpg:
Fixed in #69

PR #69, opened 2020-03-19
Update to Dotty v0.23

gzoller:
Update build to support Dotty 0.23.0-RC1


gzoller:
fixes #67

munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala
```diff
@@ -10,8 +10,8 @@ object MacroCompat {
     inline implicit def generate: Location = ${ locationImpl() }
   }
 
-  def locationImpl()(given qctx: QuoteContext): Expr[Location] = {
-    import qctx.tasty.{_, given}
+  def locationImpl()(implicit qctx: QuoteContext): Expr[Location] = {
```
SethTisue: `using` is the Scala 3 way now, I think (here and on line 24 too)

gzoller:
Actually for reasons I don’t know, they went back to implicit in some cases.   Try putting it back to using and see what happens.   It didn’t work for me and I recall reading about it.

gzoller:
Reverted to using.  I also noticed there's a yarn.lock file in the repo.  I wasn't sure if it was needed.  Presumably not--but I left it for now because it was there before.

gabro:
@gzoller `yarn.lock` is the dependecies lockfile of yarn, the javascript package manager. It's used for the website, and it's correct that is included in version control (see https://classic.yarnpkg.com/en/docs/yarn-lock/)

munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala
```diff
@@ -10,8 +10,8 @@ object MacroCompat {
     inline implicit def generate: Location = ${ locationImpl() }
   }
 
-  def locationImpl()(given qctx: QuoteContext): Expr[Location] = {
-    import qctx.tasty.{_, given}
+  def locationImpl()(using qctx: QuoteContext): Expr[Location] = {
+    import qctx.tasty.{_, given _}
```
gabro: If I'm not mistaken (by reading the release notes), this can just be

```suggestion
    import qctx.tasty._
```

munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala
```diff
@@ -21,8 +21,8 @@ object MacroCompat {
     inline implicit def generate[T](value: T): Clue[T] = ${ clueImpl('value) }
   }
 
-  def clueImpl[T:Type](value: Expr[T])(given qctx: QuoteContext): Expr[Clue[T]] = {
-    import qctx.tasty.{_, given}
+  def clueImpl[T:Type](value: Expr[T])(using qctx: QuoteContext): Expr[Clue[T]] = {
+    import qctx.tasty.{_, given _}
```
gabro: same as above

```suggestion
    import qctx.tasty._
```

.gitignore
```diff
@@ -13,6 +13,8 @@ project/boot/
 project/plugins/project/
 .bloop
 
+_site/
```
gabro: out of curiosity, where does this come from? I mean, how was `_site` generated? I've never seen it

munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala
olafurpg: I just tried applying this change and it did not compile
```
.
[error] -- [E008] Member Not Found Error: /Users/lgeirsson/dev/munit/munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala:15:28
[error] 15 |    val path = rootPosition.sourceFile.jpath.toString
[error]    |               ^^^^^^^^^^^^^^^^^^^^^^^
[error]    |               value sourceFile is not a member of qctx.tasty.Position
[error] -- [E008] Member Not Found Error: /Users/lgeirsson/dev/munit/munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala:16:33
[error] 16 |    val startLine = rootPosition.startLine + 1
[error]    |                    ^^^^^^^^^^^^^^^^^^^^^^
[error]    |value startLine is not a member of qctx.tasty.Position, but could be made available as an extension method.
[error]    |
[error]    |The following import might fix the problem:
[error]    |
[error]    |  import qctx.tasty.positionOps
[error]    |
[error] -- Error: /Users/lgeirsson/dev/munit/munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala:17:32
[error] 17 |    '{ new Location(${Expr(path)}, ${Expr(startLine)}) }
[error]    |                                ^
[error]    |ambiguous implicit arguments: both method DoubleIsLiftable in object Liftable and method LongIsLiftable in object Liftable match type quoted.Liftable[String] of an implicit parameter of method apply in object Expr
[error] -- Error: /Users/lgeirsson/dev/munit/munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala:17:52
[error] 17 |    '{ new Location(${Expr(path)}, ${Expr(startLine)}) }
[error]    |                                                    ^
[error]    |ambiguous implicit arguments: both method DoubleIsLiftable in object Liftable and method LongIsLiftable in object Liftable match type quoted.Liftable[Int] of an implicit parameter of method apply in object Expr
[error] -- [E008] Member Not Found Error: /Users/lgeirsson/dev/munit/munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala:26:23
[error] 26 |    val source = value.unseal.pos.sourceCode
[error]    |                 ^^^^^^^^^^^^
[error]    |value unseal is not a member of quoted.Expr[T], but could be made available as an extension method.
[error]    |
[error]    |The following import might fix the problem:
[error]    |
[error]    |  import qctx.tasty.QuotedExprOps.unseal
[error]    |
[error] -- Error: /Users/lgeirsson/dev/munit/munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala:28:30
[error] 28 |    '{ new Clue(${Expr(source)}, $value, ${Expr(valueType)}) }
[error]    |                              ^
[error]    |ambiguous implicit arguments: both method DoubleIsLiftable in object Liftable and method LongIsLiftable in object Liftable match type quoted.Liftable[String] of an implicit parameter of method apply in object Expr
[error] 6 errors found
[error] (munitJVM / Compile / compileIncremen
```

olafurpg:
Thank you @gzoller !

munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala
gabro: ok, thanks for trying!

PR #70, opened 2020-03-20
Relax `assertEquals` strictness about type equality.

olafurpg:
Following up from an offline conversation with @gabro

Previously, `assertEquals` only allowed comparing values of the same
type. Now, it's OK as long as the second argument is a subtype of the
first argument. The reasoning for this change is that it's common that
you want to test some generic value like `Option[Int]` obtained from a
method we're testing against an expected hardcoded value like
`Some[Int]`.

### DIFF
--- a/docs/assertions.md
+++ b/docs/assertions.md
@@ -83,19 +83,18 @@ Comparing two values of different types is a compile error.
 assertEquals(1, "")
 ```
 
-The two types must match exactly, it's a type error to compare two values even
-if one value is a subtype of the other.
+The "expected" value (second argument) must be a subtype of the "obtained" value
+(first argument). It's a type error to compare two values even if one value is a
+subtype of the other.
+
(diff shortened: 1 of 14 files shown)


docs/assertions.md
```diff
@@ -83,19 +83,18 @@ Comparing two values of different types is a compile error.
 assertEquals(1, "")
 ```
 
-The two types must match exactly, it's a type error to compare two values even
-if one value is a subtype of the other.
+The "expected" value (second argument) must be a subtype of the "obtained" value
+(first argument). It's a type error to compare two values even if one value is a
+subtype of the other.
```
gabro: is it still a type error?

olafurpg: It's not a type error, removed

olafurpg:
Yay figured out the cryptic windows failures!

PR #71, opened 2020-03-20
Add munit-scalacheck module

gabro:
This PR adds a new `munit-scalacheck` module which provides a basic level of integration with ScalaCheck via `ScalaCheckSuite`.

Example usage:

```scala
class ScalaCheckFrameworkSuite extends ScalaCheckSuite {

  property("list concatenation") {
    forAll { (l1: List[Int], l2: List[Int]) =>
      l1.size + l2.size == (l1 ::: l2).size
    }
  }

  property("squared") {
    forAll { (n: Int) =>
      scala.math.sqrt(n * n) == n
    }
  }
}
```

Example output:

```
munit.scalacheck.ScalaCheckFrameworkSuite:
OK, passed 100 tests.
  + list concatenation 0.454s
==> X munit.scalacheck.ScalaCheckFrameworkSuite.squared  0.03s java.lang.Exception: 
Falsified after 0 passed tests.
> ARG_0: -1
> ARG_0_ORIGINAL: 2147483647
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,10 +5,9 @@ val customScalaJSVersion = Option(System.getenv("SCALAJS_VERSION"))
 val scalaJSVersion = customScalaJSVersion.getOrElse("1.0.0")
 val scalaNativeVersion = "0.4.0-M2"
 def scala213 = "2.13.1"
-def scala212 = "2.12.10"
+def scala212 = "2.12.11"
 def scala211 = "2.11.12"
 def dotty = "0.23.0-RC1"
-def scalameta = "4.3.0"
 def junitVersion = "4.13"
 def gcp = "com.google.cloud" % "google-cloud-storage" % "1.103.0"
 inThisBuild(
@@ -201,8 +200,23 @@ lazy val plugin = project
     )
   )
 
-lazy val tests = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+lazy val munitScalacheck = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+  .in(file("munit-scalacheck"))
+  .settings(
+    sharedSettings,
+    crossScalaVersions := List(scala213, scala212, scala211),
+    libraryDependencies += "org.scalacheck" %%% "scalacheck" % "1.14.3"
+  )
+  .nativeConfigure(sharedNativeConfigure)
+  .nativeSettings(
+    sharedNativeSettings,
+    skip in publish := customScalaJSVersion.isDefined
+  )
+  .jsSettings(sharedJSSettings)
   .dependsOn(munit)
+
+lazy val tests = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+  .dependsOn(munit, munitScalacheck)
   .enablePlugins(BuildInfoPlugin)
   .settings(
     sharedSettings,
--- a/build.sbt
+++ b/build.sbt
@@ -202,10 +202,12 @@ lazy val plugin = project
 
 lazy val munitScalacheck = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .in(file("munit-scalacheck"))
+  .dependsOn(munit)
   .settings(
     sharedSettings,
-    crossScalaVersions := List(scala213, scala212, scala211),
-    libraryDependencies += "org.scalacheck" %%% "scalacheck" % "1.14.3"
+    crossScalaVersions := List(scala213, scala212, scala211, dotty),
+    libraryDependencies += ("org.scalacheck" %%% "scalacheck" % "1.14.3")
+      .withDottyCompat(scalaVersion.value)
   )
   .nativeConfigure(sharedNativeConfigure)
   .nativeSettings(
@@ -213,7 +215,6 @@ lazy val munitScalacheck = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     skip in publish := customScalaJSVersion.isDefined
   )
   .jsSettings(sharedJSSettings)
-  .dependsOn(munit)
 
 lazy val tests = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .dependsOn(munit, munitScalacheck)
--- a/munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
+++ b/munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
@@ -9,19 +9,25 @@ import scala.concurrent.Future
 trait ScalaCheckSuite extends FunSuite {
 
   def property(
+      name: String
+  )(body: => Prop)(implicit loc: Location): Unit = {
+    property(new TestOptions(name, Set.empty, loc))(body)
+  }
+
+  def property(
       options: TestOptions
   )(body: => Prop)(implicit loc: Location): Unit = {
     test(options)(body)
   }
 
   override def munitValueTransforms: List[ValueTransform] =
-    super.munitValueTransforms :+ scalaCheckPropTrasform
+    super.munitValueTransforms :+ scalaCheckPropTransform
 
   protected def scalaCheckTestParameters = ScalaCheckTest.Parameters.default
 
   protected def scalaCheckPrettyParameters = Pretty.defaultParams
 
-  private def scalaCheckPropTrasform: ValueTransform =
+  private def scalaCheckPropTransform: ValueTransform =
     new ValueTransform("ScalaCheck Prop", {
       case prop: Prop =>
         val result = ScalaCheckTest.check(scalaCheckTestParameters, prop)
@@ -29,11 +35,8 @@ trait ScalaCheckSuite extends FunSuite {
         if (result.passed) {
           println(prettyResult)
           Future.successful(())
-        } else
-          Future.failed {
-            val e = new Exception("\n" + prettyResult)
-            e.setStackTrace(Array.empty)
-            e
-          }
+        } else {
+          Future.failed(new ScalaCheckFailException("\n" + prettyResult))
+        }
     })
 }
--- a/munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
+++ b/munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
@@ -1,0 +1,39 @@
+package munit
+package scalacheck
+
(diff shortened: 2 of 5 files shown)


tests/shared/src/main/scala/munit/scalacheck/ScalaCheckFrameworkSuite.scala
```diff
@@ -0,0 +1,34 @@
...
+object ScalaCheckFrameworkSuite
+    extends FrameworkTest(
+      classOf[ScalaCheckFrameworkSuite],
+      s"""|==> success munit.scalacheck.ScalaCheckFrameworkSuite.list concatenation
+          |==> failure munit.scalacheck.ScalaCheckFrameworkSuite.squared -${' '}
+          |Falsified after 0 passed tests.
+          |> ARG_0: -1
+          |> ARG_0_ORIGINAL: 2147483647
+          |""".stripMargin
+    )
```
gabro: @olafurpg do you know why I don't see the exception name in the test (it should be where the `' '` is.

Also, the `println` output isn't captured here (but it appears when running `FrameworkSuite`)

munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  override def munitValueTransforms: List[ValueTransform] =
+    super.munitValueTransforms ++ List(
+      new ValueTransform("ScalaCheck Prop", {
+        case prop: Prop =>
+          val result = ScalaCheckTest.check(scalaCheckTestParameters, prop)
+          val prettyResult = Pretty.pretty(result, scalaCheckPrettyParameters)
+          if (result.passed) {
+            println(prettyResult)
```
gabro: is there a smarter thing than `println` for reporting a successful test?

munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+            Future.successful(())
+          } else
+            Future.failed {
+              val e = new Exception("\n" + prettyResult)
```
gabro: I'm throwing a generic `Exception` here, not sure there's a more appropriate way of doing this

tests/shared/src/main/scala/munit/scalacheck/ScalaCheckFrameworkSuite.scala
```diff
@@ -0,0 +1,34 @@
+package munit
+package scalacheck
+
+import org.scalacheck.Prop.forAll
+import org.scalacheck.rng.Seed
+
+class ScalaCheckFrameworkSuite extends ScalaCheckSuite {
+
+  override def scalaCheckTestParameters =
+    super.scalaCheckTestParameters.withInitialSeed(Seed(123L))
```
gabro: this is needed for making the test output stable for the failed test. It also server as a test for overriding these parameters

build.sbt
```diff
@@ -5,10 +5,9 @@ val customScalaJSVersion = Option(System.getenv("SCALAJS_VERSION"))
 val scalaJSVersion = customScalaJSVersion.getOrElse("1.0.0")
 val scalaNativeVersion = "0.4.0-M2"
 def scala213 = "2.13.1"
-def scala212 = "2.12.10"
+def scala212 = "2.12.11"
```
gabro: not strictly related, but I figured I could take the chance to update this

build.sbt
```diff
@@ -5,10 +5,9 @@ val customScalaJSVersion = Option(System.getenv("SCALAJS_VERSION"))
...
 def scala211 = "2.11.12"
 def dotty = "0.23.0-RC1"
-def scalameta = "4.3.0"
```
gabro: again, not strictly related, but this was unused

gabro:
@olafurpg I have a few doubts (see comments inline). Once we're happy with the code, I'll add some documentation

/cc @kubukoz IIRC you were doing something similar?

build.sbt
```diff
@@ -201,8 +200,23 @@ lazy val plugin = project
     )
   )
 
-lazy val tests = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+lazy val munitScalacheck = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+  .in(file("munit-scalacheck"))
+  .settings(
+    sharedSettings,
+    crossScalaVersions := List(scala213, scala212, scala211),
```
gabro: no dotty support since there are no binaries published and I couldn't make it work with `withDottyCompat` :(

gabro: ah, nevermind. The issue I had with `withDottyCompat` seems to be fixed in Dotty 0.23 🎉

tests/shared/src/main/scala/munit/scalacheck/ScalaCheckFrameworkSuite.scala
olafurpg: That's expected, `FrameworkSuite` renders custom output based on the published sbt events. This is the most reliable way I have seen so far to avoid flaky failures. There's a flag to include stdout

https://github.com/scalameta/munit/blob/37a04596254d972ee4012aa3ce95601b27fa22a4/tests/shared/src/main/scala/munit/FixtureFrameworkSuite.scala#L80

build.sbt
olafurpg: Fine to upgrade 👍

build.sbt
olafurpg: 👍

munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
olafurpg: This is OK. We have access to the sbt logger, which we could provide access to as an alternative. JUnit only supports a `didSucceed(description: Description): Unit` notification hook, a test success can't include more information.

olafurpg: The IntelliJ integration BTW captures stdout per test case

munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
```diff
@@ -0,0 +1,39 @@
+package munit
+package scalacheck
+
+import org.scalacheck.Prop
+import org.scalacheck.{Test => ScalaCheckTest}
+import org.scalacheck.util.Pretty
+import scala.concurrent.Future
+
+abstract class ScalaCheckSuite extends FunSuite {
```
olafurpg: I wonder if this should be a trait that users mix instead? It would be nice if users could mix in multiple helpers (cats integration, scalacheck, ...)

tests/shared/src/main/scala/munit/scalacheck/ScalaCheckFrameworkSuite.scala
olafurpg: 👍

olafurpg: might be worth commenting in the source instead

tests/shared/src/main/scala/munit/scalacheck/ScalaCheckFrameworkSuite.scala
gabro: 👍 thanks for the explanation

munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+    new ValueTransform("ScalaCheck Prop", {
+      case prop: Prop =>
+        val result = ScalaCheckTest.check(scalaCheckTestParameters, prop)
+        val prettyResult = Pretty.pretty(result, scalaCheckPrettyParameters)
+        if (result.passed) {
+          println(prettyResult)
+          Future.successful(())
+        } else
+          Future.failed {
+            val e = new Exception("\n" + prettyResult)
```
olafurpg: It would be good to create a new `class ScalaCheckException`. The exception type appears in JUnit test reports

gabro: 👍 makes sense

munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  override def munitValueTransforms: List[ValueTransform] =
+    super.munitValueTransforms :+ scalaCheckPropTrasform
+
+  protected def scalaCheckTestParameters = ScalaCheckTest.Parameters.default
+
+  protected def scalaCheckPrettyParameters = Pretty.defaultParams
+
+  private def scalaCheckPropTrasform: ValueTransform =
```
olafurpg:
```suggestion
  private def scalaCheckPropTransform: ValueTransform =
```

munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
olafurpg:
```scala
class ScalaCheckException(message: String) extends Exception("\n" + message) with NoStackTrace
```

munit-scalacheck/shared/src/main/scala/munit/scalacheck/ScalaCheckSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+package scalacheck
+
+import org.scalacheck.Prop
+import org.scalacheck.{Test => ScalaCheckTest}
+import org.scalacheck.util.Pretty
+import scala.concurrent.Future
+
+trait ScalaCheckSuite extends FunSuite {
+
+  def property(
```
olafurpg: Should we add a method overload that takes `property(name: String)`? For consistency with the test method. The goal is to avoid the "-language:implicitConversions` flag or `import scala.language.implicitConversions`

gabro:
By the way, I still want to add docs, but it can also be done after merge

PR #72, opened 2020-03-20
Move ScalaCheck integration into the toplevel package and add documentation

gabro:
Some docs for #71

### DIFF
--- a/docs/integrations/scalacheck.md
+++ b/docs/integrations/scalacheck.md
@@ -18,7 +18,7 @@ libraryDependencies += "org.scalameta" %% "munit-scalacheck" % "@VERSION@" % Tes
 You can then extend `ScalaCheckSuite` and write ScalaCheck property tests:
 
 ```scala mdoc
-import munit.scalacheck.ScalaCheckSuite
+import munit.ScalaCheckSuite
 import org.scalacheck.Prop._
 
 class IntegerSuite extends ScalaCheckSuite {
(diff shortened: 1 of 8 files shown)


docs/integrations/scalacheck.md
```diff
@@ -0,0 +1,142 @@
...
+}
+```
+
+## Migrating from ScalaTest
+
+ScalaTest provides two styles for writing property-based tests, which are both
+front-ends to ScalaCheck.
+
+If you are using the "ScalaCheck style" API, i.e. the `Checkers` trait,
+switching to MUnit requires very minor and mechanical changes:
```
olafurpg:
```suggestion
switching to MUnit requires minor mechanical changes:
```

docs/integrations/scalacheck.md
```diff
@@ -0,0 +1,142 @@
...
+
+```diff
+-import org.scalatest.prop.Checkers
+-import org.scalatest.FunSuite
++import munit.scalacheck.ScalaCheckSuite
```
olafurpg: Reading the docs now makes we wonder if it's a good idea to call this package `scalacheck`. It might be better to put this in the `munit` toplevel package. If users do `import munit._` then they get an unwanted `scalacheck` symbol in scope. What do you think?

gabro: I thought about it, and since scalacheck is prefixed by “org” I didn’t think it would be an issue; although I understand where you’re coming from.
I think it may make sense to either put into the toplevel package, or go the other way and do munit.integrations.scalacheck so that the intermediate name is less prone to clashes. I’m fine with either solutions

olafurpg: I’m personally a fan of large packages. There’s low risk of namespace conflicts even if the scalacheck class lives in the munit package. I can do the move if you’re ok with it.

gabro: Sure! :)

gabro: Actually, my son woke up very early, so... done!
I've done in this PR so to merge the new docs and the change together

docs/integrations/scalacheck.md
gabro: done in #76

PR #73, opened 2020-03-20
Use SCALAJS_VERSION=0.6.32 for 2.12.11 support

gabro:


gabro:
Self merging, hoping it fixes the CI

PR #77, opened 2020-03-24
Scalacheck blogpost

gabro:


website/blog/2020-02-16-async.md
```diff
@@ -23,6 +23,8 @@ The latest MUnit v0.5.0 release fixes this issue by adding support for async
 tests on all platforms meaning that the example test case above fails as
 expected after 100 milliseconds.
 
+<!-- truncate -->
```
gabro: so that this blog post doesn't take a huge amount of vertical space in https://scalameta.org/munit/blog/

build.sbt
```diff
@@ -211,7 +211,7 @@ lazy val munitScalacheck = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       .withDottyCompat(scalaVersion.value)
   )
   .jvmSettings(
-    skip in publish := customScalaJSVersion.isDefined,
+    skip in publish := customScalaJSVersion.isDefined
```
gabro: unrelated, but I think it's a leftover from the recent fix on master

website/blog/2020-03-24-scalacheck.md
```diff
@@ -0,0 +1,96 @@
...
+This is a boolean expression, which uses labels to better identify which part of
+the expression fails.
+
+For longer expressions, however, this may become inconvenient and may choose to
+use MUnit assertions instead:
+
+```scala
+property("integer identities") {
+  forAll { (n: Int) =>
+    assertEquals(n + 0, n)
```
olafurpg: Can we include a screenshot of the stdout here? Would be good to use a non-vscode terminal that supports the ANSI highlight color like iTerm.

olafurpg: I think it's OK to include the explanation here as well so that it's equivalent to the scalacheck boolean example
```scala
     assertEquals(n + 0, n, "0 is the addition identity")
```

gabro:
Thanks @olafurpg for the review. It looks much better with error output screenshots!

image

PR #78, opened 2020-03-28
Add support for Hedgehog property testing

mpilquist:
Opening this draft PR for feedback on overall API. Besides review comments, this also needs the following updates before merging:
- Hedgehog needs to publish for Scala Native
- Hedgehog PR for effectful properties needs to be merged (https://github.com/hedgehogqa/scala-hedgehog/pull/147)
- Somehow skip Scala.js 0.6 publishing of hedghog project

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -223,8 +223,31 @@ lazy val munitScalacheckJVM = munitScalacheck.jvm
 lazy val munitScalacheckJS = munitScalacheck.js
 lazy val munitScalacheckNative = munitScalacheck.native
 
+lazy val munitHedgehog = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+  .in(file("munit-hedgehog"))
+  .dependsOn(munit)
+  .settings(
+    moduleName := "munit-hedgehog",
+    sharedSettings,
+    crossScalaVersions := List(scala213, scala212, scala211, dotty),
+    resolvers += "bintray-scala-hedgehog".at("https://dl.bintray.com/hedgehogqa/scala-hedgehog"),
+    libraryDependencies += ("qa.hedgehog" %% "hedgehog-runner" % "97854199ef795a5dfba15478fd9abe66035ddea2").withDottyCompat(scalaVersion.value)
+  )
+  .jvmSettings(
+    skip in publish := customScalaJSVersion.isDefined
+  )
+  .nativeConfigure(sharedNativeConfigure)
+  .nativeSettings(
+    sharedNativeSettings,
+    skip in publish := customScalaJSVersion.isDefined
+  )
+  .jsSettings(sharedJSSettings)
+lazy val munitHedgehogJVM = munitHedgehog.jvm
+lazy val munitHedgehogJS = munitHedgehog.js
+lazy val munitHedgehogNative = munitHedgehog.native
+
 lazy val tests = crossProject(JSPlatform, JVMPlatform, NativePlatform)
-  .dependsOn(munit, munitScalacheck)
+  .dependsOn(munit, munitScalacheck, munitHedgehog)
   .enablePlugins(BuildInfoPlugin)
   .settings(
     sharedSettings,
--- a/build.sbt
+++ b/build.sbt
@@ -230,8 +230,11 @@ lazy val munitHedgehog = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     moduleName := "munit-hedgehog",
     sharedSettings,
     crossScalaVersions := List(scala213, scala212, scala211, dotty),
-    resolvers += "bintray-scala-hedgehog".at("https://dl.bintray.com/hedgehogqa/scala-hedgehog"),
-    libraryDependencies += ("qa.hedgehog" %% "hedgehog-runner" % "97854199ef795a5dfba15478fd9abe66035ddea2").withDottyCompat(scalaVersion.value)
+    resolvers += "bintray-scala-hedgehog".at(
+      "https://dl.bintray.com/hedgehogqa/scala-hedgehog"
+    ),
+    libraryDependencies += ("qa.hedgehog" %% "hedgehog-runner" % "97854199ef795a5dfba15478fd9abe66035ddea2")
+      .withDottyCompat(scalaVersion.value)
   )
   .jvmSettings(
     skip in publish := customScalaJSVersion.isDefined
--- a/build.sbt
+++ b/build.sbt
@@ -233,7 +233,7 @@ lazy val munitHedgehog = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     resolvers += "bintray-scala-hedgehog".at(
       "https://dl.bintray.com/hedgehogqa/scala-hedgehog"
     ),
-    libraryDependencies += ("qa.hedgehog" %% "hedgehog-runner" % "97854199ef795a5dfba15478fd9abe66035ddea2")
+    libraryDependencies += ("qa.hedgehog" %%% "hedgehog-runner" % "97854199ef795a5dfba15478fd9abe66035ddea2")
       .withDottyCompat(scalaVersion.value)
   )
   .jvmSettings(
--- a/munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
+++ b/munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
@@ -4,6 +4,8 @@ import hedgehog._
 import hedgehog.core.{PropertyConfig, PropertyT, Seed}
 import hedgehog.predef.Monad
 
+import munit.internal.Compat._
+
 /**
   * Provides the ability to write property based tests using the Hedgehog library.
   * 
@@ -31,12 +33,12 @@ trait HedgehogSuite extends FunSuite {
   lazy val propertySeed: Long = System.currentTimeMillis()
 
   def property(
-      name: String,
+      name: String
   )(prop: PropertyT[Result])(implicit loc: Location): Unit =
     property(name, propertyConfig)(prop)
 
   def property(
-      options: TestOptions,
+      options: TestOptions
   )(prop: PropertyT[Result])(implicit loc: Location): Unit =
     property(options, propertyConfig)(prop)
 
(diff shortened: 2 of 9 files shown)


munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
```diff
@@ -0,0 +1,115 @@
...
+    property(name, propertyConfig)(prop)
+
+  def property(
+      options: TestOptions
+  )(prop: PropertyT[Result])(implicit loc: Location): Unit =
+    property(options, propertyConfig)(prop)
+
+  def property(
+      name: String,
+      config: PropertyConfig
```
olafurpg: To reduce the number of method overloads, would it make sense to pass the property config via tags instead?
```scala
property("name".tag(customPropertyConfig))
```

build.sbt
```diff
@@ -223,8 +223,34 @@ lazy val munitScalacheckJVM = munitScalacheck.jvm
...
 
+lazy val munitHedgehog = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+  .in(file("munit-hedgehog"))
+  .dependsOn(munit)
+  .settings(
+    moduleName := "munit-hedgehog",
+    sharedSettings,
+    crossScalaVersions := List(scala213, scala212, scala211, dotty),
+    resolvers += "bintray-scala-hedgehog".at(
+      "https://dl.bintray.com/hedgehogqa/scala-hedgehog"
```
olafurpg: It would be nice to have a hedgehog release on Maven Central.

munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
mpilquist: I looked in to that but note doing so would require the ability to convert property config to/from String, unless I'm misunderstanding.

build.sbt
mpilquist: Agreed, there's an open issue for it and lead maintainer has been working it but not sure how close it is to being merged.

munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
mpilquist: BTW, I used `Conversion` to avoid the need for importing `language.implicitConversions` when working with Dotty. Maybe we should do the same for `Conversion[String, TestOptions]`?

olafurpg: It's possible to create a tag that contains custom values like this
```scala
case class HedgehogConfig(config: PropertyConfig) extends Tag("HedgehogConfig")
```
See https://scalameta.org/munit/docs/tests.html#customize-evaluation-of-tests-with-tags

I didn't know about the `Conversion` trick to avoid the import. That's handy. However, I think it's good to keep the `name: String` method overload to make the sources more approachable for Scala beginners

mpilquist: Ah, cool, I'll update the overloads.

Re: `Conversion`, I'm getting warnings with munit 0.7.1 when calling `test("name".ignore)`. I was wondering if we could use `Conversion` to avoid such warnings. I'll open a separate issue on that.

olafurpg: Please do!

munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
```diff
@@ -0,0 +1,115 @@
...
+      options: TestOptions,
+      config: PropertyConfig
+  )(prop: PropertyT[F[Result]])(implicit loc: Location): Unit =
+    test(options)(checkF(prop, config, propertySeed))
+
+  /**
+   * Checks the supplied `Property[Result]`, throwing a `HedgehogFailException`
+   * if the property was falsified.
+   */
+  def check(
```
gabro, line 83: would it make sense to make the `check` methods private?

munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
```diff
@@ -0,0 +1,115 @@
...
+ */
+trait HedgehogSuite extends FunSuite {
+
+  /**
+   * Provides the default configuration for running property tests.
+   * This can be overriden to change the default or can be changed on
+   * a test-by-test basis by using an overload of `property` and `propertyF`.
+   */
+  lazy val propertyConfig: PropertyConfig = PropertyConfig.default
+  lazy val propertySeed: Long = System.currentTimeMillis()
```
gabro, line 33: Not a huge deal, but I would like to standardize these with respect to the scalacheck module. A few differences I can spot:

- we use `protected` in `ScalaCheckSuite` (maybe redundant, but I think it makes sense)
- we use `def` for the config, to allow overriding with anything (`def`,  `val`, `lazy val`) instead of forcing to use `val` in the override
- we prefix the variables using `scalaCheck` as in `scalaCheckTestParameters`. Since these modules are defined as `trait` I wanted to maintain some kind of namespace to avoid potential clashes when we add more modules that can be mixed and matched.

I'm fine to standardize either way, as long as it's consistent.

What do you think @mpilquist?

gabro, line 33: Also, does it make sense to memoize the seed? IIUC in ScalaCheck the seed is generated fresh for each prop

gabro: Minor, but would it make sense to use

```suggestion
  lazy val propertySeed: Seed = Seed.fromTime()
```

instead?

See https://github.com/hedgehogqa/scala-hedgehog/blob/master/core/shared/src/main/scala/hedgehog/core/Seed.scala#L48-L49

tests/shared/src/main/scala/munit/ScalaCheckFrameworkSuite.scala
```diff
@@ -7,7 +7,7 @@ class ScalaCheckFrameworkSuite extends ScalaCheckSuite {
 
   // NOTE(gabro): this is needed for making the test output stable for the failed test below.
   // It also serves as a test for overriding these parameters.
-  override def scalaCheckTestParameters =
+  override def scalaCheckTestParameters: org.scalacheck.Test.Parameters =
```
gabro, line 10: was this required by Scalafix? I must have missed it

munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
mpilquist: I made it public b/c in scodec-bits, I had some cases where I wanted to call `check` multiple times in a single test. That's not really necessary though so I'm happy with this being private.

munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
mpilquist: 👍 on `protected` and prefixing names. Re: def versus val, I used lazy val in order to guarantee the same seed is used in all tests in a suite. I'm okay with using a def instead though. No strong opinion.

gabro: > def versus val, I used lazy val in order to guarantee the same seed is used in all tests in a suite

👍 makes sense if that's the intended semantics. Does the same reasoning apply to the config?

mpilquist: Yeah I'd like to use `Seed` but need to figure out a way to convert an arbitrary seed to/from a string for reporting.

tests/shared/src/main/scala/munit/ScalaCheckFrameworkSuite.scala
gabro: anyway, if you rebase on master that's gone now

mpilquist: Yep

munit-hedgehog/shared/src/main/scala/munit/HedgehogSuite.scala
gabro: I ended up using `String` myself in `ScalaCheckSuite`, since it's then easier for the user to set it when trying to reproduce a failure.

See the updated docs at https://scalameta.org/munit/docs/integrations/scalacheck.html#reproducing-a-property-failure

tests/shared/src/main/scala/munit/HedgehogFrameworkSuite.scala
```diff
@@ -42,8 +42,7 @@ class HedgehogFrameworkSuite extends HedgehogSuite {
   }
 
   property(
-    "custom config",
-    config = propertyConfig.copy(testLimit = SuccessCount(1000))
+    "custom config".tag(HedgehogConfig(hedgehogPropertyConfig.copy(testLimit = SuccessCount(1000))))
```
olafurpg, line 45: This is quite a mouthful, I wonder if we want to add custom tags for the most common configuration options? Just brainstorming, it would be nice if the user could do something like this instead 🤔 

```suggestion
    "custom config".tag(TestLimit(1000))
```

gabro: If we go down this route I would still namespace the tags

```suggestion
    "custom config".tag(HedgehogTestLimit(1000))
```

olafurpg: This is totally optional BTW, just a suggestion. I'm happy to merge this PR as is. I believe we can update this PR to skip projects where the cross-build is missing

mpilquist: I agree it's too verbose as-is. I'll update like you suggested. I think I'll add support for passing a seed as a tag as well.

I can also remove the `propertyF` stuff for now (until Hedgehog merges support for that).

One note about tags - figuring out the supported tags is not discoverable via auto-completion. We could address that via extension methods - what do you think?

olafurpg:
Closing this PR since it's been inactive for a while. Don't hesitate to reopen if you want to pick this up again!

PR #90, opened 2020-03-29
Print property seeds and add docs about reproducing test failures

gabro:
Previously we would not print seeds when a property would fail, so it would be impossible to deterministically reproduce a failure.
Now we print the seed when a property fails and we added docs to explain how to reproduce failures

### DIFF
--- a/docs/integrations/scalacheck.md
+++ b/docs/integrations/scalacheck.md
@@ -137,13 +137,13 @@ You can reproduce this failure by adding this to your suite:
 
 To reproduce the failure you can follow the suggestion to fix the seed:
 
-```scala
-class MySuite extends ScalaCheckSuite {
+```diff
+ class MySuite extends ScalaCheckSuite {
 
-  override val scalaCheckInitialSeed = "CTH6hXj8ViScMmsO78-k4_RytXHPK_wSJYNH2h4dCpB="
++  override val scalaCheckInitialSeed = "CTH6hXj8ViScMmsO78-k4_RytXHPK_wSJYNH2h4dCpB="
 
-  // ...
-}
+   // ...
+ }
 ```
 
 Re-running the test will now fail deterministically, which allows you to work on
--- a/docs/integrations/scalacheck.md
+++ b/docs/integrations/scalacheck.md
@@ -117,6 +117,33 @@ class IntegerSuite extends ScalaCheckSuite {
 }
 ```
 
+## Reproducing a property failure
+
+In some cases, you may find that ScalaCheck properties fail
+non-deterministically. This can happen due to the randomness of the input values
+for each test run.
+
+When this happens, you can deterministally reproduce a property failure by using
+its seed. MUnit prints the property seed whenever a failure occurs in the form
+of:
+
+```
+Failing seed: CTH6hXj8ViScMmsO78-k4_RytXHPK_wSJYNH2h4dCpB=
+```
+
+To reproduce the exact same failure, you can override the test parameters and
+provide the offending seed as the initial seed:
+
+```scala
+  override def scalaCheckTestParameters =
+    super.scalaCheckTestParameters.withInitialSeed(
+      "CTH6hXj8ViScMmsO78-k4_RytXHPK_wSJYNH2h4dCpB="
+    )
+```
+
+Re-running the test will now fail deterministically, which allows you to work on
+an appropriate fix without worrying about randomness.
+
 ## Migrating from ScalaTest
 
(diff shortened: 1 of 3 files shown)


munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
```diff
@@ -34,21 +35,37 @@ trait ScalaCheckSuite extends FunSuite {
...
     })
 
-  private def propToTry(prop: Prop, testLocation: Location): Try[Unit] = {
+  private def propToTry(prop: Prop, test: Test): Try[Unit] = {
     import ScalaCheckTest._
-    val result = check(scalaCheckTestParameters, prop)
-    def renderResult(r: Result) =
-      Pretty.pretty(r, scalaCheckPrettyParameters)
+    val seed = scalaCheckTestParameters.initialSeed.getOrElse(Seed.random())
+    val result = check(scalaCheckTestParameters, prop.useSeed(test.name, seed))
```
gabro: The ScalaCheck runner uses `prop.viewSeed(name, seed)` but that causes the prop to use an hardcoded `println` to output the seed, which would look quite messy since it appears before the test failure in the output.

Instead, we determine the seed in the same way ScalaCheck does (getting it from the parameters or falling back to a random one) and inject it in the property using `useSeed` (which is what `propertyWithSeed` does in ScalaCheck)

Now we're free to print the seed ourselves, which looks much better and more consistent with the rest of the failure message.

munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
```diff
@@ -34,21 +35,37 @@ trait ScalaCheckSuite extends FunSuite {
 
   protected def scalaCheckPrettyParameters = Pretty.defaultParams
 
+  // Allow specifying an initial seed in its Base64 encoded form
+  implicit class XtensionScalaCheckTestParameters(
+      params: ScalaCheckTest.Parameters
+  ) {
+    def withInitialSeed(encodedSeed: String): ScalaCheckTest.Parameters =
+      params.withInitialSeed(Seed.fromBase64(encodedSeed).get)
+  }
```
gabro: ScalaCheck does something analogous in `propertyWithSeed`, and I think it makes sense (since we print the Base64 encoded form in test failures.

docs/integrations/scalacheck.md
```diff
@@ -117,6 +117,33 @@ class IntegerSuite extends ScalaCheckSuite {
...
+
+```
+Failing seed: CTH6hXj8ViScMmsO78-k4_RytXHPK_wSJYNH2h4dCpB=
+```
+
+To reproduce the exact same failure, you can override the test parameters and
+provide the offending seed as the initial seed:
+
+```scala
+  override def scalaCheckTestParameters =
```
olafurpg: What do you think about making it possible to override the default seed with a string? The format of the output could also be 

override def scalaCheckSeed = “Ceirgejskfh”

So that users can copy paste it

gabro:
ScalaCheck itself uses base64 encoded seeds in the output. See an example here https://gist.github.com/non/aeef5824b3f681b9cfc141437b16b014

You can still use Long if you prefer, but I preferred to stay consistent with the default ScalaCheck runner

olafurpg:
Using a Base64 encoded string is fine.

docs/integrations/scalacheck.md
gabro: implemented, take a look!

docs/integrations/scalacheck.md
```diff
@@ -117,6 +117,38 @@ class IntegerSuite extends ScalaCheckSuite {
...
+Failing seed: CTH6hXj8ViScMmsO78-k4_RytXHPK_wSJYNH2h4dCpB=
+You can reproduce this failure by adding this to your suite:
+
+  override val scalaCheckInitialSeed = "CTH6hXj8ViScMmsO78-k4_RytXHPK_wSJYNH2h4dCpB="
+
+```
+
+To reproduce the failure you can follow the suggestion to fix the seed:
+
+```scala
```
olafurpg, line 140: In these cases I normally use the `diff` language to highlight what's changed

```diff

 class MySuite extends ScalaCheckSuite {
+   override val scalaCheckInitialSeed = "CTH6hXj8ViScMmsO78-k4_RytXHPK_wSJYNH2h4dCpB="
    // ...
 }
```

gabro: nice idea 👍 done!

PR #103, opened 2020-04-13
Add file layout utilities

gabro:
Port `FileLayout` utilities from Metals (https://github.com/scalameta/metals/blob/e2320b5b4d7f365c6064ccc3ecb2a105b579a109/tests/unit/src/main/scala/tests/FileLayout.scala#L9)

I had to add a few nio methods to make this work in JS.

Also, I had to change the split regex to make it work in Scala Native (due to https://github.com/scala-native/scala-native/issues/620)


gabro:
@olafurpg I'll add some docs before merging once we're happy with the code

munit/js/src/main/scala/java/nio/file/Files.scala
```diff
@@ -24,4 +31,59 @@ object Files {
...
+
+  def openOptionsToFlag(options: Seq[OpenOption]): Option[String] = {
+    import StandardOpenOption._
+    val combinations = List(
+      Set[OpenOption](WRITE, READ, CREATE) -> "w+",
+      Set[OpenOption](WRITE, READ, CREATE_NEW) -> "wx+",
+      Set[OpenOption](WRITE, CREATE) -> "w",
+      Set[OpenOption](WRITE, CREATE_NEW) -> "wx"
+      // TODO(gabro): all the other combinations
+    )
```
gabro, line 82: I _think_ this makes sense: Java allows combining arbitrary options, while Node lists out the flags. I can add the remaining ones before merging

munit/shared/src/main/scala/munit/FileLayout.scala
```diff
@@ -0,0 +1,75 @@
...
+   * @param layout
+   * @return the file layout as a map of path -> content
+   * @throws IllegalArgumentException if the layout is formally invalid
+   */
+  protected def parseFileLayout(layout: String): Map[String, String] =
+    if (!layout.trim.isEmpty) {
+      val lines = layout.replaceAllLiterally("\r\n", "\n")
+      lines
+      // lines starting with '/' are file paths, but lines starting with '//' are comments
+        .split("(\n/[^/])")
```
gabro, line 25: here's where I've changed the regex.
It was 

```suggestion
        .split("(?=\n/[^/])")
```

before, but Scala Native implementation does not support lookaheads.

I've adjusted line 27 to not remove the leading "\n" (since it's dropped by `split` now) and it seems to work as expected

munit/js/src/main/scala/java/nio/file/Files.scala
```diff
@@ -24,4 +31,59 @@ object Files {
...
+  def deleteIfExists(path: Path): Boolean = JSIO.inNode {
+    if (JSIO.exists(path.toString)) {
+      JSFs.unlinkSync(path.toString)
+      true
+    } else {
+      false
+    }
+  }
+
+  def write(path: Path, bytes: Array[Byte], options: OpenOption*): Path =
```
olafurpg, line 57: The JDK boot classpath takes precedence over regular classpath entries during compilation, while regular classpath entries are used during linking time (since they have sjsir files). Therefore, varargs need to use the same erasure signature as Java varargs
```suggestion
  def write(path: Path, bytes: Array[Byte], options: Array[OpenOption]): Path =
```

munit/js/src/main/scala/java/nio/file/Files.scala
```diff
@@ -24,4 +31,59 @@ object Files {
...
+      }
+      JSFs.writeFileSync(
+        path.toString(),
+        Uint8Array.from(data),
+        js.Dynamic.literal(flag = openOptionsToFlag(options).orUndefined)
+      )
+      path
+    }
+
+  def openOptionsToFlag(options: Seq[OpenOption]): Option[String] = {
```
olafurpg: For compatibility with JDK APIs
```suggestion
  private def openOptionsToFlag(options: Seq[OpenOption]): Option[String] = {
```

munit/shared/src/main/scala/munit/FileLayout.scala
```diff
@@ -0,0 +1,75 @@
+package munit
+
+import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.StandardOpenOption
+import java.nio.file.Path
+
+trait FileLayout {
```
olafurpg, line 9: What do you think about turning this into an object?
```scala
object FileLayouts {
  def parse(..)
  def write(..)
}
```
This could help improve the completion experience in IDEs and also avoids adding `{parse,write}FileLayout` to the `FunSuite` scope 🤔

munit/js/src/main/scala/java/nio/file/Files.scala
gabro: I've gone ahead and added all cases

munit/shared/src/main/scala/munit/FileLayout.scala
```diff
@@ -0,0 +1,75 @@
...
+  protected def fileLayoutDefaultTempDirName = "file-layout"
+
+  /**
+   * Parse a file layout from string
+   *
+   * @param layout
+   * @return the file layout as a map of path -> content
+   * @throws IllegalArgumentException if the layout is formally invalid
+   */
+  protected def parseFileLayout(layout: String): Map[String, String] =
```
olafurpg, line 20: Should this be a public method? We don't have an equivalent method to this in `scala.meta.testkit.StringFS` https://github.com/scalameta/scalameta/blob/3d66f6c561d26886dda508aff0c314d65761df68/scalameta/testkit/src/main/scala/scala/meta/testkit/StringFS.scala

Instead, we have something like `parse(path: Path): String` which recursively walks a directory and renders a layout.

munit/js/src/main/scala/java/nio/file/Files.scala
gabro: 👍 makes sense (although is it expected that this works in tests?)

munit/shared/src/main/scala/munit/FileLayout.scala
gabro: I initially had it implemented like this, but then changed my mind. I thought it would be nice to avoid an import to use `FileLayout` methods so I decided to make it a trait and mix it in, to make it as usable as `Assertions`

We could also do

```scala
trait FileLayout {
  object FileLayout {
    def parse(...)
    def write(...)
  }
}
```

or something like that, what do you think?

munit/shared/src/main/scala/munit/FileLayout.scala
gabro: Are you proposing to introduce `parse(path: Path): String` and making `parse(layout: String): Map[String, String]` private?

I think parsing the layout is useful when looking for specific patterns in a file. Recursively walking the directory will pick-up all the files, which may not be needed and be unnecessarily expensive.

For reference, I'm using it in  
https://github.com/buildo/retro/blob/master/tapiro/core/src/test/scala/io/buildo/tapiro/TapiroSuite.scala#L127-L131

munit/js/src/main/scala/java/nio/file/Files.scala
olafurpg: No need to change it if it works in the tests

munit/shared/src/main/scala/munit/FileLayout.scala
olafurpg: I'm not convinced this utility is used enough to merit inclusion in the `FunSuite` scope, unlike assertions. I think it's also desirable to have access to this object outside of `FunSuite`. We can have a `val FileLayout = munit.FileLayout` forwarder inside `FunSuite` but I think we're entering a slippery slope then and end up adding every new feature into the `FunSuite` scope

munit/shared/src/main/scala/munit/FileLayout.scala
olafurpg: We can keep `parse(layout: String): Map[String, String]` public but how about renaming it into `parseLayout`? That leaves it open to later implement `parsePath`

gabro:
@olafurpg ok, I've reworked this a bit:

- `FileLayout` is now a class with a `write` method and a few useful methods in the companion object (`fromString` and `write`, which creates a `FileLayout` and writes it to disk)
- the API meant to be used is `munit.FileLayout.write` or `munit.FileLayout.fromString`, so no more trait mixed in `FunSuite`: I agree that if we start polluting the `FunSuite` is becomes a slippery slope
- I would like to add `fromPath`, but it requires porting some more cross-target file IO for JS and native, so I'd defer it to a future PR

gabro:
Added some docs too (it probably needs another pass, but it's a start to see whether we like the examples and the structure)

docs/file-layout.md
```diff
@@ -0,0 +1,98 @@
...
+import munit.FunSuite
+import munit.FileLayout
+
+class MyCliSuite extends FunSuite {
+
+  test("credentials-from-file") {
+    val authPath = ".myCli/auth"
+    val expectedToken = "token=abcdef123456"
+    val root = FileLayout.write(
+      s"""|/${authPath}
```
olafurpg, line 49: I feel like we can come up with a better motivating example. In this specific example we're not taking much advantage of the multiple-line syntax, it might even be clearer if anything with a `Map[String, String]` data structure.

The biggest value in file layouts is when you have large literal strings and minimal interpolated variables, ideally with multiple files in nested directories (like the `D.scala` and `e.json` example below).

docs/file-layout.md
```diff
@@ -0,0 +1,98 @@
...
+}
+
+class MyCodeGenSuite extends FunSuite {
+  test("generated-files") {
+    val layout = """..."""
+    val root = MyCodeGen.generateFiles()
+    FileLayout.fromString(layout).files.map {
+      case (path, expectedContent) =>
+        val actualContent = new String(Files.readAllBytes(root.resolve(path)))
+        assertNoDiff(actualContent, expectedContent)
```
olafurpg, line 94: If I was writing a test case for a code generator that generates multiple files then I would instead use `FileLayout.fromPath(..)` to render a large multiline layout string and then use a single `assertNoDiff(..)`. The benefit of this approach is that we aggregate errors for multiple files and that the printed stripMargin string can be copy-pasted into the test when the expected behavior changes. If we assert against individual files then we need to run the test multiple times to get all the errors.

Issue #108, opened 2020-04-20
Unexpected blocked test execution


jvican:
When writing a test case that uses `assertNoDiff` with two big strings, munit gets blocked and the test doesn't finish.
  
After having a closer look with `jstack`, it looks like the test is stuck at `newCapacity`, which is triggered by `AnsiColors.filterAnsi`.

```
"pool-1-thread-1" #19 prio=5 os_prio=31 tid=0x00007f8cfea70800 nid=0x6603 runnable [0x00007000102e9000]
   java.lang.Thread.State: RUNNABLE
	at java.lang.AbstractStringBuilder.newCapacity(AbstractStringBuilder.java:154)
	at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:125)
	at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:448)
	at java.lang.StringBuilder.append(StringBuilder.java:136)
	at munit.internal.console.AnsiColors$.filterAnsi(AnsiColors.scala:35)
	at munit.internal.difflib.Diff.<init>(Diff.scala:9)
	at munit.internal.difflib.Diffs$.assertNoDiff(Diffs.scala:20)
	at munit.Assertions.$anonfun$assertNoDiff$1(Assertions.scala:53)
	at munit.Assertions$$Lambda$194/744865283.apply$mcZ$sp(Unknown Source)
	at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.scala:17)
	at munit.internal.console.StackTraces$.dropInside(StackTraces.scala:10)
	at munit.Assertions.assertNoDiff(Assertions.scala:53)
	at munit.Assertions.assertNoDiff$(Assertions.scala:47)
	at pulsar.testing.BaseSpec.assertNoDiff(BaseSuite.scala:450)
	at pulsar.NbformatSpec.$anonfun$checkNbformatRoundTrip$6(NbformatSpec.scala:16)
	at pulsar.NbformatSpec.$anonfun$checkNbformatRoundTrip$6$adapted(NbformatSpec.scala:14)
	at pulsar.NbformatSpec$$Lambda$122/2086796509.apply(Unknown Source)
	at monix.eval.Task$Map.apply(Task.scala:4541)
	at monix.eval.Task$Map.apply(Task.scala:4537)
	at monix.eval.internal.TaskRunLoop$.startFuture(TaskRunLoop.scala:550)
[...]

PR #109, opened 2020-04-20
Use `StringBuilder` in `AnsiColors`

jvican:
Fixes https://github.com/scalameta/munit/issues/108

### DIFF
--- a/munit/shared/src/main/scala/munit/internal/console/AnsiColors.scala
+++ b/munit/shared/src/main/scala/munit/internal/console/AnsiColors.scala
@@ -23,7 +23,7 @@ object AnsiColors {
     if (s == null) {
       null
     } else {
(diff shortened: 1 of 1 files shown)


jvican:
I confirm this PR fixes #108. I'm not sure how I could reproduce this bug because the inputs for `assertNoDiff` are quite esoteric (lots of characters, base64, weird encodings, etc) and I can't share them. If you have an idea for a test, let me know @olafurpg 
  
Sorry for not submitting a PR from my fork, I realized I was pushing to the munit repository too late

olafurpg:
I pushed one commit just to make it closer to a line-by-line translation of https://github.com/scalameta/munit/blob/53e8379cedc0d1364a2569a05fbf1b63e1f280cc/junit-interface/src/main/java/munit/internal/junitinterface/Ansi.java#L31

PR #117, opened 2020-04-28
Add support for Dotty 0.23.0

olafurpg:
Previously, we only built against the latest Dotty release candidate. Now, we
build against both the release candidate and the latest stable release.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,8 +7,8 @@ val scalaNativeVersion = "0.4.0-M2"
 def scala213 = "2.13.2"
 def scala212 = "2.12.11"
 def scala211 = "2.11.12"
-def dotty = "0.24.0-RC1"
-def dottyLegacy = "0.23.0"
+def dottyNext = "0.24.0-RC1"
+def dottyStable = "0.23.0"
 def junitVersion = "4.13"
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -8,6 +8,7 @@ def scala213 = "2.13.2"
 def scala212 = "2.12.11"
 def scala211 = "2.11.12"
 def dotty = "0.24.0-RC1"
+def dottyLegacy = "0.23.0"
```
gabro, line 11:
```suggestion
def dottyNext = "0.24.0-RC1"
def dottyStable = "0.23.0"
```

build.sbt
```diff
@@ -56,7 +57,8 @@ addCommandAlias(
 )
 val isPreScala213 = Set[Option[(Long, Long)]](Some((2, 11)), Some((2, 12)))
 val scala2Versions = List(scala211, scala212, scala213)
-val scalaVersions = scala2Versions ++ List(dotty)
+val scala3Versions = List(dotty, dottyLegacy)
```
gabro, line 60:
```suggestion
val scala3Versions = List(dottyStable, dottyNext)
```

PR #124, opened 2020-05-07
Regenerate default seed at every prop iteration

### LINKED ISSUE #118: ScalaCheckSuite unexpected behavior
It seems that the ScalaCheckSuite tests are not generating different values during the `forAll` iterations for types that are not collections or Strings (like `Int`, `Double` or even `UUID`). Tests like the below
```scala
final class IntGeneratorsSuite extends ScalaCheckSuite {

  val ints: mutable.Set[Int] = mutable.Set.empty

  property("generating int") {
    forAll { i: Int =>
      ints.add(i)
      println(ints)
      true
    }
  }

  test("generated int are not all the same") {
    assert(ints.size > 1)
  }
}
```
fail with the current master.  and print like
```
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
HashSet(-66888203)
[...]
```
This might be related to the fact that an initial seed is always set [here](https://github.com/scalameta/munit/blob/master/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala#L56) and scalacheck copies that into the `Gen`'s initial seed and keeps using it to create values (see `genPrms` [here](https://github.com/typelevel/scalacheck/blob/f64f0cc3f21d9ba34848853181eaca1ebd60e39e/src/main/scala/org/scalacheck/Test.scala#L410) ,`forAllShrink` [here](https://github.com/typelevel/scalacheck/blob/09a243146f70a6f0b3ac02bde6013cd25675a314/src/main/scala/org/scalacheck/Prop.scala#L760) and `startSeed` [here](https://github.com/typelevel/scalacheck/blob/09a243146f70a6f0b3ac02bde6013cd25675a314/src/main/scala/org/scalacheck/Prop.scala#L517).
I'm not sure about what a possible solution could be. I played a bit with the code but I couldn't find any way to set an initial seed in the `Test`'s params and get different values as I was expecting.

gabro:
Fixes #118 

This is extreme subtle, but it seems to work now

### DIFF
--- a/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
+++ b/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
@@ -49,11 +49,14 @@ trait ScalaCheckSuite extends FunSuite {
 
   private def propToTry(prop: Prop, test: Test): Try[Unit] = {
     import ScalaCheckTest._
-    val seed =
+    def seed =
       scalaCheckTestParameters.initialSeed.getOrElse(
         Seed.fromBase64(scalaCheckInitialSeed).get
       )
-    val result = check(scalaCheckTestParameters, prop.useSeed(test.name, seed))
+    val result = check(
+      scalaCheckTestParameters,
+      Prop(genParams => prop(genParams.withInitialSeed(seed)))
+    )
     def renderResult(r: Result) = {
       val resultMessage = Pretty.pretty(r, scalaCheckPrettyParameters)
       if (r.passed) {
--- a/tests/shared/src/test/scala/munit/ScalaCheckSeedSuite.scala
+++ b/tests/shared/src/test/scala/munit/ScalaCheckSeedSuite.scala
@@ -1,0 +1,21 @@
+package munit
+
+import scala.collection.mutable
+import org.scalacheck.Prop.forAll
+
+// Regression test for https://github.com/scalameta/munit/issues/118
+final class ScalaCheckSeedSuite extends ScalaCheckSuite {
+
+  private val ints = mutable.Set.empty[Int]
+
+  property("generating int") {
+    forAll { (i: Int) =>
+      ints.add(i)
+      true
+    }
+  }
+
+  test("generated int are not all the same") {
+    assert(ints.size > 1)
+  }
+}


munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
```diff
@@ -49,11 +49,14 @@ trait ScalaCheckSuite extends FunSuite {
...
-    val seed =
+    def seed =
       scalaCheckTestParameters.initialSeed.getOrElse(
         Seed.fromBase64(scalaCheckInitialSeed).get
       )
-    val result = check(scalaCheckTestParameters, prop.useSeed(test.name, seed))
+    val result = check(
+      scalaCheckTestParameters,
+      Prop(genParams => prop(genParams.withInitialSeed(seed)))
+    )
```
gabro, line 59: this way if the user hasn't overridden the default seed, we recompute it in the `Prop` lambda (which is what `viewSeed` does in vanilla ScalaCheck)

barambani:
Thanks for the fix. This version works for all the tests I had.

gabro:
Thanks for confirming and thanks again for the detailed bug report. Having a reproduction test case has been really helpful in tracking this down

olafurpg:
It’s possible to use println in framework suites just to check that each iteration has a unique value.

gabro:
Wooos, I actually had a test suite I forgot to commit 😅 doing it tomorrow

gabro:
Done now @olafurpg

olafurpg:
This fix has been released as v0.7.6. Thank you @barambani for reporting!

barambani:
Brilliant, Thank you both.

PR #129, opened 2020-05-13
Preserve stack trace of unexpected exceptions in ScalaCheckSuite

### LINKED ISSUE #128: fatal errors thrown in scalacheck mode lose stacks
see this discussion:

https://gitter.im/scalameta/munit?at=5eb9dc2310d5ec031aff33d9

you can repro it with a simple property:
```scala
  property("hide my stacks") {
    Prop.forAll { i: Int =>
      if (i > 0) throw new StackOverflowError()
      else Prop(true)
    }
  }
```

olafurpg:
Previously, stack traces from unexpected exceptions would be swallowed
in ScalaCheckSuite. Now, the stack trace is displayed making it easier
to troubleshoot a test failure.

Unexpected exceptions look like this now
Screenshot 2020-05-13 at 09 04 16

Fixes #128

### DIFF
--- a/tests/shared/src/main/scala/munit/ScalaCheckExceptionFrameworkSuite.scala
+++ b/tests/shared/src/main/scala/munit/ScalaCheckExceptionFrameworkSuite.scala
@@ -1,0 +1,70 @@
+package munit
+
+import org.scalacheck.Prop
+
+class ScalaCheckExceptionFrameworkSuite extends ScalaCheckSuite {
+  override val scalaCheckInitialSeed =
+    "9WNU_CSZAQwiiPWDlHs4NWI-knIDEKCsgGqdhZgNKnB="
+
+  property("hide my stacks") {
+    Prop.forAll { (i: Int) =>
+      throw new StackOverflowError()
+      Prop(true)
+    }
+  }
+}
+
+// NOTE(olafur): extracted this object into a class to avoid the following error with Dotty
+// ==> X munit.FrameworkSuite.initializationError  0.002s java.lang.VerifyError: Bad type on operand stack
+// Exception Details:
+//   Location:
+//     munit/ScalaCheckExceptionFrameworkSuite$.<init>()V @19: invokedynamic
+//   Reason:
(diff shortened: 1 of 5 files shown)


tests/shared/src/main/scala/munit/ScalaCheckExceptionFrameworkSuite.scala
```diff
@@ -0,0 +1,69 @@
...
+//     locals: { uninitializedThis, 'java/lang/Class', 'java/lang/String' }
+//     stack: { uninitializedThis }
+//   Bytecode:
+//     0x0000000: 1210 4cb2 0015 b200 1a12 1cb6 0020 b600
+//     0x0000010: 234d 2aba 0037 0000 4eb2 003c b600 403a
+//     0x0000020: 04b2 003c b600 443a 05b2 003c b600 483a
+//     0x0000030: 062a 2b2c 1904 1905 1906 2dbb 004a 5912
+//     0x0000040: 4c10 33b7 004f b700 522a b300 54b1
+//
+// FrameworkSuite.<init>(FrameworkSuite.scala:30)
```
gabro, line 36: did you have a chance to report this?

olafurpg: I didn't report it since it takes more time to minimize the issue than I have right now.

gabro: :+1: cool, just wondering if we could refer to an issue

gabro: maybe @smarter is interested in this, since MUnit is in the community build already

smarter: If there's a bug, please open an issue, I can't keep track of everything  by myself.

olafurpg: I would open an issue if I had a minimized example but the code here relies on non-published code that we use only for tests.

Issue #93, opened 2020-03-31
Feature Req: Describe section


gzoller:
I know the main philosophy of munit is to keep it lean and simple, so I understand if this feature is not part of that picture--but I'll ask anyway.

There are so many great highlighting features in munit I don't miss my old standby test framework, scalatest a lot.  One feature scalatest did have that was nice to have was a describe{} section that allowed tests in a file to be sectioned and intended.

For example:
```scala
describe(
    "---------------------------\n:  Scala Primitive Tests  :\n---------------------------"
  ) {
      describe("+++ Positive Tests +++") {
           // several tests here
      }
      describe("+++ Positive Tests +++") {
           // several tests here
      }
}
```
This would produce a nice sectioned display:
```
---------------------------
:  Scala Primitive Tests  :
---------------------------
+++ Positive Tests +++
   -- test output
   -- test output
+++ Positive Tests +++
   -- test output
   -- test output
```

This feature is a nice way to section and format groups of tests.  Describe is only an output/formatting thing--it doesn't "do" anything in terms of the tests themselves.

olafurpg:
Thank you for reporting! I am concerned this is a slippery slope to take. Can you elaborate on the difference between `describe(..)` and `println(..)` with some ANSI colors? Is the following a valid workaround?

```scala
import munit.internal.consol
def describe(message: String)(thunk: => Unit): Unit = {
  println(Console.GREEN + messsage + Console.RESET)
  thunk
}
```
The only difference from what I can see is that the "test output" lines will not be indented

gzoller:
This is very close.  I can live w/o the indent, and I do understand the slippery slope.

When I do this, with this decribe() method:
```scala
describe("---A---") {
   test("Foo") {...}
   test("Bar") {...}
describe("---B---") {
   test("Blather") {...}
   test("Baz") {...}
}
```
I get this output
```
---A---
---B---
  + Foo
  + Bar
  + Blather
  + Baz
```
I was hoping for:
```
---A---
  + Foo
  + Bar
---B---
  + Blather
  + Baz
```
Can this be done with the facilities already in munit?

gzoller:
Did some experiments implementing munit.Printer and munit.Printable, mainly just to see if anything passed through there.  In my tests, nothing passed thru print() (nothing was output), so maybe I haven't wired it up correctly?  Don't know if this is the right direction, but it is what I tried next.
```scala
class Collections extends munit.FunSuite with munit.Printer {
[...]

PR #133, opened 2020-05-24
Add support for nested tests via JUnit Description children

olafurpg:
Previously, MUnit only supported declaring tests as a flat list. Now,
  users can organize test cases as a nested tree of test like this.
```scala
class NestedSuite extends munit.FunSuite {
  describe("primitives") {
    describe("positive") {
      test("positive1") {}
      test("positive2") {}
    }
    describe("negative") {
      test("negative1") {}
      test("negative2") {}
    }
  }
}
```

The rendered output of this indents the nested tests like this:
```
munit.NestedSuite:
  primitives 0.054s
    positive 0.001s
      + positive1 0.0s
      + positive2 0.0s
    negative 0.0s
      + negative1 0.0s
      + negative2 0.0s
```

Related issue #93 

cc/ @mpilquist @gzoller 

I'm personally on the fence about introducing this behavior because it's not clear how nested tests interact with

* filtering capabilities: it's technically not possible to run only a nested test, you must run all the parents as well
* fixtures: for example, do we trigger `beforeEach` for every node in the test tree? 

Users can already group related tests via test suites, which provides a better UX for filtering (tab completions in sbt) and it's more self-explanatory how fixtures work (you have hooks before/after each test case and for each test suite)

TODOs

* [ ] Scala.js/Native support
* [ ] Make sure IntelliJ renders nested tests correctly, not as a flat list

Screenshot 2020-05-24 at 12 38 46

### DIFF
--- a/junit-interface/src/main/java/munit/internal/junitinterface/Describe.java
+++ b/junit-interface/src/main/java/munit/internal/junitinterface/Describe.java
@@ -1,0 +1,5 @@
+package munit.internal.junitinterface;
+
+import java.lang.annotation.Annotation;
+
+public interface Describe {}
--- a/junit-interface/src/main/java/munit/internal/junitinterface/EventDispatcher.java
+++ b/junit-interface/src/main/java/munit/internal/junitinterface/EventDispatcher.java
@@ -95,7 +95,7 @@ final class EventDispatcher extends RunListener
     uncapture(true);
     postIfFirst(new ErrorEvent(failure, Status.Failure) {
       void logTo(RichLogger logger) {
-        logger.error( settings.buildTestResult(Status.Failure) +ansiName+" "+ durationSuffix() + " " + ansiMsg, error);
+        logger.error(settings.buildTestResult(failure.getDescription(), Status.Failure) +ansiName+" "+ durationSuffix() + " " + ansiMsg, error);
       }
     });
   }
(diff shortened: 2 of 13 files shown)


gabro:
It's an interesting experiment, however I'm on the fence too: the complexity it adds when interplaying with other features such as filtering or fixtures doesn’t seem justified by the benefits it brings.

I don’t feel strongly about it, but I personally lean towards leaving this out

olafurpg:
Thanks for the input! There might be space for a middle way where we only support something like `describe()` but not arbitrarily nested tests. That would solve the issues around filtering and fixtures while keeping the ability to group tests.

barambani:
I have recently migrated a big set of tests to Munit (I'm still doing it) and I admit that sometimes being able to group the cases would have made my life easier. In particular having a way to `describe` a group of tests would have the benefit, in my case, of reducing the length and the amount of text I have to write in every `text` or `property`. But I wouldn't sacrifice the simplicity of Munit or confuse its fundamental semantics for it. I think the `describe` option without nesting is an excellent compromise.

mpilquist:
I like it! Arbitrary nesting of `describe` with `test` nodes at the leaves resolves the issues with filtering and lifecycles.

gabro:
> There might be space for a middle way where we only support something like `describe()` but not arbitrarily nested test

I like the idea 👍

olafurpg:
Sounds good. I'm 👍 with the idea of adding support for something like `describe()` that is internally implemented on top of nested JUnit descriptions (for example, so that it renders correctly as a nested tree in IntelliJ). Closing this PR since it implements a bit more general solution (arbitrarily nested tests)

PR #165, opened 2020-07-17
add ScalaCheckSuite#include

nevillelyh:
Not sure if this fits here, but it allows us to include tests from an existing ScalaCheck `Properties`, e.g. discipline laws
https://github.com/typelevel/discipline/blob/327ab715dcff99f756fbff2ff75ada2df7d80112/core/src/main/scala/org/typelevel/discipline/Laws.scala#L83

### DIFF
--- a/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
+++ b/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
@@ -1,6 +1,6 @@
 package munit
 
-import org.scalacheck.Prop
+import org.scalacheck.{Prop, Properties}
 import org.scalacheck.{Test => ScalaCheckTest}
 import org.scalacheck.util.Pretty
 import org.scalacheck.rng.Seed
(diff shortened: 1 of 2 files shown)


gabro:
Thank you @nevillelyh!

I'm a bit wary of this change since the ScalaCheck integration tries to do the bare minimum to allow using ScalaCheck with MUnit.

While `include` it's indeed useful in the specific use case, I'm afraid it would be a slippery slope if other use cases start popping up and the API surface risks to get bigger and bigger.

Given that `include` it's a tiny utility function used to glue together existing ScalaCheck props definitions to MUnit suites, I think this could live in "userland" and we could advertise it in the documentation as a recommended solution if you want to port over existing suites. This would align well with our strategy for dealing with ScalaTest migrations (we explain how do it, but MUnit does not have any features in that sense).

What do you all think?

nevillelyh:
Yeah I agree. It's tiny and I worked around it by keeping the method in my "userland".
Also 👍  on keeping everything small & bare minimal :)

PR #166, opened 2020-07-17
doc: flesh out the conversion to ScalaTest docs a good deal

Kazark:
* Link back to relevant information elsewhere in the docs
* Add tactical tips
* Explain differences in `build.sbt` between MUnit and ScalaTest
* Add some structure to the now-copious information

Hopefully this will encourage adoption!

Note that I was not able to test locally; I could not get the build running. I want to to test that the headers looked okay, and that the links worked. However, I was fairly careful about putting in the links, so I expect it should be fine. If you are at all nervous about it, you may want to render it and see if it is okay.

### DIFF
--- a/docs/scalatest.md
+++ b/docs/scalatest.md
@@ -3,6 +3,34 @@ id: scalatest
 title: Coming from ScalaTest
 ---
 
+# Initiating the Conversion
+
+Begin by ensuring that you have
+[added a dependency](./getting-started.html#quick-start) on MUnit. If you are
+using SBT, don't neglect to explicitly add MUnit as a test framework:
+
+```diff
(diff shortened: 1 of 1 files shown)


olafurpg:
Thank you for this pr! It may take a couple weeks for me to look into this since I’ll be away on holidays

Kazark:
@olafurpg Thanks for letting me know. I've taken these notes now while I'm fresh, so I'm in no rush. Enjoy your holidays!

olafurpg (approved):
This documentation is fantastic. Very clear and easy to follow along. Thank you for this contribution! I'm gonna merge even if the CI is red, I can't restart the jobs for some reason but I don't see a reason why the build would fail since there are no code changes.

PR #192, opened 2020-09-04
Assert equals doubles

OlivierBlanvillain:


munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -120,6 +120,40 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
...
+      obtained: Double,
+      expected: Double,
+      delta: Double,
+      clue: => Any
+  )(implicit loc: Location): Unit = {
+    StackTraces.dropInside {
+      val exactlyTheSame = java.lang.Double.compare(expected, obtained) == 0
+      val almostTheSame = Math.abs(expected - obtained) <= delta
+      if (!exactlyTheSame && !almostTheSame) {
+        fail(munitPrint(clue))
```
olafurpg, line 152: Can we include the `obtained` and `expected` values in the error message?

tests/shared/src/test/scala/munit/DoubleAssertionsSuite.scala
```diff
@@ -0,0 +1,126 @@
...
+
+class DoubleAssertionsSuite extends BaseSuite {
+  test("Assert Equals NaN Fails") {
+    try {
+      assertEquals(1.234, Double.NaN, 0.0)
+      fail("unexpected")
+    } catch {
+      case _: FailException =>
+    }
+  }
```
olafurpg, line 11: You can use `.fail` to assert a test fails
```suggestion
  test("Assert Equals NaN Fails".fail) {
    assertEquals(1.234, Double.NaN, 0.0)
  }
```

tests/shared/src/test/scala/munit/DoubleAssertionsSuite.scala
```diff
@@ -0,0 +1,126 @@
+package munit
+
+class DoubleAssertionsSuite extends BaseSuite {
```
olafurpg, line 3: You may want to convert this test suite into a "framework suite", which makes it easier to assert success/fail results. Example https://github.com/scalameta/munit/blob/af3931c3f18d9c79fd72f06fa7734e03c3d140b3/tests/shared/src/main/scala/munit/FailFrameworkSuite.scala

olafurpg (changes requested):
Thank you for this contribution! This is a good idea 👍 A few suggestions

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -123,13 +123,27 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
...
+      obtained: Double,
+      expected: Double,
+      delta: Double
+  )(implicit loc: Location): Unit = {
+    val defaultClue = "values are not the same"
+    assertEquals(obtained, expected, delta, defaultClue)
+  }
+
+  /**
+   * Asserts that two doubles or floats are equal to within a positive delta.
```
olafurpg, line 138: I'm not the right person to ask about scaladoc, I almost never use it 😅

tests/shared/src/test/scala/munit/DoubleAssertionsSuite.scala
olafurpg: You'll need to manually register the test here https://github.com/scalameta/munit/blob/af3931c3f18d9c79fd72f06fa7734e03c3d140b3/tests/shared/src/test/scala/munit/FrameworkSuite.scala#L29

OlivierBlanvillain:
My PR was nonsense since both `assertEquals` and `fail` throw a `FailException`. But it looks like I'm not the only who made that mistake https://github.com/junit-team/junit4/pull/1673 :sweat_smile:

I'm not so sure about the usefulness of last commit, is this what you meant?

Also I'll be completely offline for the next ~10 days, but if the PR needs more work I promise I will come back to afterwards :pray:

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -120,6 +120,40 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
     }
   }
 
+  /**
+   * Asserts that two doubles or floats are equal to within a positive delta.
+   * If the expected value is infinity then the delta value is ignored.
+   * NaNs are considered equal: assertEquals(Double.NaN, Double.NaN, *) passes.
+   */
+  def assertEquals(
```
olafurpg, line 128: What do you think about renaming this method?

```suggestion
  def assertEqualsWithDelta(
```

It would be nice to avoid method overloads, currently there is only one `assertEquals`.

OlivierBlanvillain: How about `assertEqualsDouble` (and maybe `assertEqualsFloat`)? The reasoning being that you would basically never want to check the equality of two floating point numbers without using a delta.

olafurpg: Sounds good to me.

olafurpg (approved):
LGTM 👍 Thank you @OlivierBlanvillain 

I pushed a commit to remove the framework suite. Framework suites are only helpful if you want to assert the failure message

olafurpg:
I triggered a 0.7.13 release with this fix https://github.com/scalameta/munit/releases/tag/v0.7.13

PR #195, opened 2020-09-08
Report accurate ScalaCheck seed on failing test

nevillelyh:
Failing seed reporting seems to be incorrect when `scalaCheckInitialSeed` is not overridden. In that case `def seed` would use `def scalaCheckInitialSeed: String = Seed.random().toBase64` and be different every time. Hence the 2 seeds seen here are actually not the ones causing test failure.

https://travis-ci.org/github/spotify/magnolify/jobs/708925430#L369

```
Failing seed: z9omj4mAsT7BcJQoh95CcofqurdkO_jIAGEWFPx7xZA=
You can reproduce this failure by adding the following override to your suite:
  override val scalaCheckInitialSeed = "5E72c0hWZ5ioOvDIj2udNT9KuVs1mGz8EBcV-3zyKgH="
```

### DIFF
--- a/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
+++ b/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
@@ -49,13 +49,17 @@ trait ScalaCheckSuite extends FunSuite {
 
   private def propToTry(prop: Prop, test: Test): Try[Unit] = {
     import ScalaCheckTest._
-    def seed =
+    def makeSeed() =
       scalaCheckTestParameters.initialSeed.getOrElse(
         Seed.fromBase64(scalaCheckInitialSeed).get
       )
+    var seed: Seed = null
     val result = check(
       scalaCheckTestParameters,
-      Prop(genParams => prop(genParams.withInitialSeed(seed)))
+      Prop { genParams =>
+        seed = makeSeed()
+        prop(genParams.withInitialSeed(seed))
+      }
     )
     def renderResult(r: Result): String = {
       val resultMessage = Pretty.pretty(r, scalaCheckPrettyParameters)
--- a/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
+++ b/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
@@ -53,12 +53,14 @@ trait ScalaCheckSuite extends FunSuite {
       scalaCheckTestParameters.initialSeed.getOrElse(
         Seed.fromBase64(scalaCheckInitialSeed).get
       )
-    var seed: Seed = null
+    val initialSeed = makeSeed()
+    var seed: Seed = initialSeed
     val result = check(
       scalaCheckTestParameters,
       Prop { genParams =>
-        seed = makeSeed()
-        prop(genParams.withInitialSeed(seed))
+        val r = prop(genParams.withInitialSeed(seed))
+        seed = seed.next
+        r
       }
     )
     def renderResult(r: Result): String = {
@@ -66,10 +68,10 @@ trait ScalaCheckSuite extends FunSuite {
       if (r.passed) {
         resultMessage
       } else {
-        val seedMessage = s"""|Failing seed: ${seed.toBase64}
+        val seedMessage = s"""|Failing seed: ${initialSeed.toBase64}
                               |You can reproduce this failure by adding the following override to your suite:
                               |
-                              |  override val scalaCheckInitialSeed = "${seed.toBase64}"
+                              |  override val scalaCheckInitialSeed = "${initialSeed.toBase64}"
                               |""".stripMargin
         seedMessage + "\n" + resultMessage
       }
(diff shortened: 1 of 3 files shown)


olafurpg (approved):
Thank you for this contribution! LGTM 👍

Any thoughts @gabro ?

gabro (approved):
Looks great, thanks @nevillelyh for tracking this down and fixing it!

nevillelyh:
Not sure how to test though, probably need some variation of `FrameworkTest`, extract the seed from error message & verify.

nevillelyh:
Actually DO NOT MERGE. Let me verify if this actually works as expected first.

olafurpg:
I am happy to merge without a test case (although it would be nice to have a test 😄 )

nevillelyh:
Think this actually broke another case. It's hard to verify since the test is non-deterministic :sob:. At least let me verify some basic scenarios.

munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
```diff
@@ -49,23 +49,29 @@ trait ScalaCheckSuite extends FunSuite {
 
   private def propToTry(prop: Prop, test: Test): Try[Unit] = {
     import ScalaCheckTest._
-    def seed =
+    def makeSeed() =
       scalaCheckTestParameters.initialSeed.getOrElse(
         Seed.fromBase64(scalaCheckInitialSeed).get
       )
+    val initialSeed = makeSeed()
```
nevillelyh, line 56: This should be the actual "initial seed", one that's used to generate seeds for all test runs (default 100) of the same property.

munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
```diff
@@ -49,23 +49,29 @@ trait ScalaCheckSuite extends FunSuite {
...
+    var seed: Seed = initialSeed
     val result = check(
       scalaCheckTestParameters,
-      Prop(genParams => prop(genParams.withInitialSeed(seed)))
+      Prop { genParams =>
+        val r = prop(genParams.withInitialSeed(seed))
```
nevillelyh, line 61: A single test run here. And advance seed to the next.

munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
```diff
@@ -49,23 +49,29 @@ trait ScalaCheckSuite extends FunSuite {
...
+        r
+      }
     )
     def renderResult(r: Result): String = {
       val resultMessage = Pretty.pretty(r, scalaCheckPrettyParameters)
       if (r.passed) {
         resultMessage
       } else {
-        val seedMessage = s"""|Failing seed: ${seed.toBase64}
+        val seedMessage = s"""|Failing seed: ${initialSeed.toBase64}
```
nevillelyh, line 71: If a test failed, say 67th out of 100, we still report the 1st seed in case the test is not pure.

tests/shared/src/test/scala/munit/ScalaCheckSeedSuite.scala
```diff
@@ -15,6 +15,25 @@ final class ScalaCheckSeedSuite extends ScalaCheckSuite {
...
+  test("generated int are not all the same") {
+    println(ints)
+    assert(ints.size > 1)
+  }
+}
+
+final class ScalaCheckInitialSeedSuite extends ScalaCheckSuite {
+
+  // initial seed should be used for the first out of 100 `minSuccessfulTests` only
+  override val scalaCheckInitialSeed = "9SohH7wEYXCdXK4b9yM2d6TKIN2jBFcBs4QBta-2yTD="
```
nevillelyh, line 27: This test checks that when initial seed is overridden, we still generate different input per test run instead of reusing the same.

nevillelyh (commented):
Added some comments. Seems to work now at least as far as my understanding of ScalaCheck & requirement goes.

tests/shared/src/test/scala/munit/ScalaCheckSeedSuite.scala
```diff
@@ -15,6 +15,26 @@ final class ScalaCheckSeedSuite extends ScalaCheckSuite {
     }
   }
 
+  test("generated int are not all the same") {
+    println(ints)
```
olafurpg, line 19: Remove?

tests/shared/src/test/scala/munit/ScalaCheckSeedSuite.scala
```diff
@@ -15,6 +15,26 @@ final class ScalaCheckSeedSuite extends ScalaCheckSuite {
...
+    assert(ints.size > 1)
+  }
+}
+
+final class ScalaCheckInitialSeedSuite extends ScalaCheckSuite {
```
olafurpg, line 24: Can we move this to a separate file? I normally try to organize one test suite per file.

olafurpg (changes requested):
@nevillelyh Thank you for double checking! A few minor comments, otherwise this looks ready to go

tests/shared/src/test/scala/munit/ScalaCheckSeedSuite.scala
nevillelyh: Fixed

tests/shared/src/test/scala/munit/ScalaCheckSeedSuite.scala
nevillelyh: Fixed

olafurpg (approved):
LGTM 👍 I'll let @gabro do the merge

gabro (approved):
Thanks again!

PR #212, opened 2020-10-04
MUnitRunner.filter should mutate state as in other standard JUnitRunners

unkarjedy:
Accordingly to org.junit.runner.manipulation.Filterable.filter documentation:
>Remove tests that don't pass the parameter filter

This allows IntelliJ "Rerun Failed Tests" feature work properly (don't report tests which are not run as "skipped" (with grey color))
https://youtrack.jetbrains.com/issue/SCL-17360#focus=Comments-27-4414969.0-0

### DIFF
--- a/munit/shared/src/main/scala/munit/MUnitRunner.scala
+++ b/munit/shared/src/main/scala/munit/MUnitRunner.scala
@@ -32,19 +32,28 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
 
   def this(cls: Class[_ <: Suite]) =
     this(MUnitRunner.ensureEligibleConstructor(cls), () => cls.newInstance())
+
   val suite: Suite = newInstance()
-  private val suiteDescription = Description.createSuiteDescription(cls)
+
   private implicit val ec: ExecutionContext = suite.munitExecutionContext
-  @volatile private var filter: Filter = Filter.ALL
+
   @volatile private var settings: Settings = Settings.defaults()
   @volatile private var suiteAborted: Boolean = false
-  val descriptions: mutable.Map[suite.Test, Description] =
+
+  private val descriptions: mutable.Map[suite.Test, Description] =
     mutable.Map.empty[suite.Test, Description]
-  val testNames: mutable.Set[String] = mutable.Set.empty[String]
-  lazy val munitTests: Seq[suite.Test] = suite.munitTests()
+  private val testNames: mutable.Set[String] =
+    mutable.Set.empty[String]
+
+  private lazy val munitTests: mutable.ArrayBuffer[suite.Test] =
+    mutable.ArrayBuffer[suite.Test](suite.munitTests(): _*)
 
   override def filter(filter: Filter): Unit = {
-    this.filter = filter
+    val newTests = munitTests.filter { t =>
+      filter.shouldRun(createTestDescription(t))
+    }
+    munitTests.clear()
+    munitTests ++= newTests
   }
   override def configure(settings: Settings): Unit = {
     this.settings = settings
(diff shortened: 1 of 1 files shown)


munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -32,19 +32,28 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
+
+  private val descriptions: mutable.Map[suite.Test, Description] =
     mutable.Map.empty[suite.Test, Description]
-  val testNames: mutable.Set[String] = mutable.Set.empty[String]
-  lazy val munitTests: Seq[suite.Test] = suite.munitTests()
+  private val testNames: mutable.Set[String] =
+    mutable.Set.empty[String]
+
+  private lazy val munitTests: mutable.ArrayBuffer[suite.Test] =
+    mutable.ArrayBuffer[suite.Test](suite.munitTests(): _*)
```
olafurpg, line 49: What is the idiomatic JUnit way to report a failure in the `filter()` method? I suspect it's OK to throw an exception but just want to be sure.

The method call `suite.munitTests()` has a chance of crashing since it calls user-land code. Previously, we only accessed `munitTests` when running tests where we could notify about errors in `RunNotifier` but we can't do that from the `filter()` method. We may want to move the `StackTraces.dropOutside()` call below into this line here.

olafurpg (commented):
Thank you for this contribution! I am really excited to get improved MUnit support in IntelliJ 😄 

Just one small question, otherwise looks good to me

munit/shared/src/main/scala/munit/MUnitRunner.scala
unkarjedy: I'm not JUnit guru, but AFAIU ideally `filter` shouldn't fail.
From what I can see it's implied that `filter` can ony throw `NoTestsRemainException` if no tests remain:
```java
public interface Filterable {
    void filter(Filter filter) throws NoTestsRemainException;
}
```

Maybe let's replace lazy val version
```scala
lazy val munitTests: mutable.ArrayBuffer[suite.Test]
``` 
with 
```scala
private var munitTests: immutable.Seq[suite.Test] =
  StackTraces.dropOutside {
    suite.munitTests().toList
  }

override def filter(filter: Filter): Unit = {
  val newTests = munitTests.filter { t =>
    filter.shouldRun(createTestDescription(t))
  }
  munitTests = newTests
}
```
and leave a single place where initialization error can occur?

olafurpg: IIRC, the user gets a fairly cryptic runtime error if the class fails to initialize. The error message was clearer when reported via `RunNotifier`. What I propose we do instead is

* wrap `suite.munitTests()` inside try/catch
* if try/catch fails, store exception inside some var and fallback to an empty list of tests
* if the exception is non-empty, report an error via `RunNotifier`

unkarjedy: image
image

Hm, can you elaborate on the "cryptic runtime error if the class fails to initialize"?
I get similar output for both variants (lazy val and var) ins sbt:

```scala
import org.junit.Assert

class MyMUnitSuite extends munit.FunSuite {
  test("test success") {
    assertEquals(1, 1)
  }
  test("test munit failure") {
    this.assertEquals(1, 2)
  }
  test("test junit failure") {
    Assert.assertEquals(1, 2)
  }

  throw new RuntimeException("Some exception in constructor")

  test(s"test junit failure ${1/0}") {
    Assert.assertEquals(1, 2)
  }
}
```

unkarjedy: Maybe it's somehow related to js/native support?
If so, could you please provide some example sbt multiplatform project with munit+js/native?

olafurpg: My bad, I should have validated this was actually an issue (I only recalled dealing with some related issues). Thank you for checking!

olafurpg (approved):
LGTM 👍 Thank @unkarjedy for this contribution!

PR #223, opened 2020-10-18
Switch to Scala 3 universal equality for `assertEquals()`

### LINKED ISSUE #189: Wrong order of arguments for assertEquals

olafurpg:
Previously, MUnit had a subtyping constraint on `assertEquals(a, b)` so that it would fail to compile if `a` was not a subtype of `b`. This was a suboptimal solution because the compile error messages could become cryptic in some cases. Now, MUnit uses the new `Eql[A, B]` multiversal equality in Scala 3 (http://dotty.epfl.ch/docs/reference/contextual/multiversal-equality.html). For Scala 2, we drop the subtyping constraint so that it's no longer a compile error to compare unrelated types. If the values are different, then users will see a runtime error instead.

Fixes #189

### DIFF
--- a/munit/shared/src/main/scala-2/munit/internal/EqualityCompat.scala
+++ b/munit/shared/src/main/scala-2/munit/internal/EqualityCompat.scala
@@ -1,0 +1,13 @@
+package munit.internal
+
+trait EqualityCompat {
+
+  /**
+   * A Scala 2 "polyfill" for the upcoming multiversal equality in Scala 3
+   *
+   * http://dotty.epfl.ch/docs/reference/contextual/multiversal-equality.html
+   */
+  sealed abstract class Eql[A, B]
+  implicit def universalEqualityForScala2[A, B]: Eql[A, B] =
+    null.asInstanceOf[Eql[A, B]]
+}
--- a/build.sbt
+++ b/build.sbt
@@ -131,28 +131,8 @@ lazy val junit = project
 lazy val munit = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(
     sharedSettings,
-    unmanagedSourceDirectories.in(Compile) ++= {
-      val root = baseDirectory.in(ThisBuild).value / "munit"
-      val base = root / "shared" / "src" / "main"
-      val result = mutable.ListBuffer.empty[File]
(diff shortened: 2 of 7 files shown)


munit/shared/src/main/scala-2/munit/internal/EqualityCompat.scala
```diff
@@ -0,0 +1,13 @@
+package munit.internal
+
+trait EqualityCompat {
+
+  /**
+   * A Scala 2 "polyfill" for the upcoming multiversal equality in Scala 3
+   *
+   * http://dotty.epfl.ch/docs/reference/contextual/multiversal-equality.html
+   */
+  sealed abstract class Eql[A, B]
```
olafurpg, line 10: cc/ @OlivierBlanvillain @smarter what is the recommended `Eql` "polyfill" for Scala 2 in cross-built library APIs?

gabro: This is the implementation in the docs

```
package scala
import annotation.implicitNotFound

@implicitNotFound("Values of types ${L} and ${R} cannot be compared with == or !=")
sealed trait Eql[-L, -R]

object Eql {
  object derived extends Eql[Any, Any]
}
```

We can probably copy paste this, and add `implicit` to the `derived` object. We can also probably drop the implicitNotFound annotation for Scala 2 (since it will never be triggered)

gabro (approved):
👍 just a suggestion about Eql for scala 2

olafurpg:
I should probably have marked this PR as a draft. I’m still experimenting with different alternatives to see if we can support custom equality type classes like cats.Eq

munit/shared/src/main/scala-2/munit/internal/EqualityCompat.scala
smarter: You're the first person to do it so there's no recommendation yet :). Personally I think I would use separate source files for Scala 2 and 3 in the places where I would want to use Eql.

olafurpg: I went with a separate `Compare[A, B]` type-class that I use in both Scala 2 and Scala 3, and in Scala 3 I provide default instances based on `Eql[A, B]`.

olafurpg:
Superseded by #225

PR #225, opened 2020-10-19
Introduce "strict equality" mode for `assertEquals()` and friends.

### LINKED ISSUE #189: Wrong order of arguments for assertEquals

olafurpg:
Previously, MUnit had a subtyping constraint on `assertEquals(a, b)` so
that it would fail to compile if `a` was not a subtype of `b`. This was
a suboptimal solution because the compile error messages could become
cryptic in some cases. Additionally, this API didn't integrate with
other libaries like Cats that has its own `cats.Eq[A,B]` type-class.

Now, MUnit uses a new `munit.Compare[A,B]` type-class for comparing
values of different types. By default, MUnit provides a "universal"
instance that permits comparison between all types and uses the built-in
`==` method. Users can optionally enable "strict equality" by adding the
compiler option `"-Xmacro-settings.munit.strictEquality"` in Scala 2.
In Scala 3, we use the `Eql[A, B]` type-classes instead to determine
type equality (per http://dotty.epfl.ch/docs/reference/contextual/multiversal-equality.html)

Note: the first few commits are unrelated to this feature, I cleaned up
a few small quirks I noticed while working on this feature.

Fixes #189


olafurpg:
Big thanks to @gabro who helped a lot with getting this PR into shape 🙏

munit/shared/src/main/scala-2/munit/internal/MacroCompat.scala
```diff
@@ -1,14 +1,90 @@
...
+    }
+  }
+
+  def compareImpl[A: c.WeakTypeTag, B: c.WeakTypeTag](
+      c: Context
+  ): c.Expr[Compare[A, B]] = {
+    import c.universe._
+    if (!isStrictEqualityEnabled(c)) {
+      reify(Compare.defaultCompare[A, B])
+    } else {
```
gabro, line 56: minor concern: this doesn't cover the case in which a user wants to customize the diff but use universal equality. Right now it's not a big issue, but it may not scale if we make more customizations available via the `Compare` typeclass

munit/shared/src/main/scala-2/munit/internal/MacroCompat.scala
```diff
@@ -1,14 +1,90 @@
...
+      else if (lineContent.contains("assertNotEquals"))
+        solutions += "use assertNotEquals[Any, Any](...) if you think it's OK to compare these types at runtime"
+
+      solutions += s"""disable strict equality by removing the compiler option "-Xmacro-settings:${strictEqualityFlag}""""
+
+      val allSolutions = solutions.zipWithIndex
+        .map { case (msg, i) => s"  Alternative ${i + 1}: $msg" }
+        .mkString("\n")
+
+      c.error(
```
gabro, line 75: this is probably me not know macros at all, but doesn't this macro always report an error when strict equality is enabled? Where is the code path that reifies the correct `Compare[A, B]` instance?

munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala
```diff
@@ -1,11 +1,17 @@
...
+import munit.Compare
 import munit.Location
 import scala.quoted._
 
 object MacroCompat {
 
+  trait CompareMacro {
+    implicit def defaultCompareImplicit[A, B](using eql: Eql[A, B]): Compare[A, B] =
+      Compare.defaultCompare[A, B]
+  }
```
gabro, line 13: Not it's idiomatic, but I think in Scala 3 we can do:

```suggestion
  trait CompareMacro {
    given [A, B](using eql: Eql[A, B]) as Compare[A, B] = Compare.defaultCompare[A, B]
  }
```

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -67,20 +54,25 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
...
+   * Asserts that two elements are not equal according to the `Compare[A, B]` type-class.
+   *
+   * By default, uses `==` to compare values.
+   */
   def assertNotEquals[A, B](
       obtained: A,
       expected: B,
       clue: => Any = "values are the same"
-  )(implicit loc: Location, ev: A =:= B): Unit = {
+  )(implicit loc: Location, ev: Compare[A, B]): Unit = {
```
gabro, line 73:
```suggestion
  )(implicit loc: Location, compare: Compare[A, B]): Unit = {
```

for consistency with `assertEquals`

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -67,20 +54,25 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
...
     StackTraces.dropInside {
-      if (obtained == expected) {
+      if (ev.isEqual(obtained, expected)) {
```
gabro, line 75:
```suggestion
      if (compare.isEqual(obtained, expected)) {
```

munit/shared/src/main/scala/munit/Compare.scala
```diff
@@ -0,0 +1,90 @@
...
+      expected
+    )(loc)
+  }
+
+}
+
+object Compare extends MacroCompat.CompareMacro {
+  private val anyEquality: Compare[Any, Any] = new Compare[Any, Any] {
+    def isEqual(a: Any, b: Any): Boolean = a == b
+  }
```
gabro, line 87: This is me having a thing for SAMs, but I think this may actually work 😅

```suggestion
  private val anyEquality: Compare[Any, Any] = _ == _
```

gabro (changes requested):
Looking good! 😍 Just left some comments here and there

munit/shared/src/main/scala-2/munit/internal/MacroCompat.scala
olafurpg: If you have a custom implicit `Compare[A, B]` in scope then you won't hit on this macro in the first place. My first attempt to implement this macro had infinite recursion because I tried to fallback to `implicitly[Compare[A, B]]`, which triggered another macro expansion 😅 There's no way to do "pick the next implicit in scope", as far as I understand

munit/shared/src/main/scala-2/munit/internal/MacroCompat.scala
olafurpg: This macro is never called if you have a custom `Compare[A, B]` in scope. I added a comment
```
      // NOTE: report compile error because the user has enabled strict equality
      // and implicit search still picked the default instance in the
      // low-priority `CompareMacro` trait.
```

munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala
olafurpg: TIL, I updated the implementation accordingly!

munit/shared/src/main/scala/munit/Assertions.scala
olafurpg: 👍 Done

munit/shared/src/main/scala/munit/Compare.scala
olafurpg: You and your love for SAM 😜 It's so much nicer with SAM though, I'm starting to like SAM more and more

olafurpg: Done

olafurpg (commented):
Thank you for the review @gabro !

munit/shared/src/main/scala-2/munit/internal/MacroCompat.scala
gabro: Ahh, makes sense. Thanks!

olafurpg:
Copy-pasting the last commit message explaining the latest approach, which I think comes out quite nicely. Kudos to @gabro for spotting this approach was possible!

---


Drop strict equality, allow comparison between supertypes/subtypes

This is a fourth attempt at improving strict equality in MUnit
`assertEquals()` assertions.

* First attempt (current release version): require second argument to be
  a supertype of the first argument. This has the flaw that the compile
  error message is cryptic and that the ordering of the arguments affects
  compilation.
* Second attempt: use `Eql[A, B]` in Scala 3 and allow comparing any
  types in Scala 2. This has the flaw that it's a regression in some
  cases for Scala 2 users and that `Eql[A, B]` is not really usable
  in its current form, see related discussion
  https://contributors.scala-lang.org/t/should-multiversal-equality-provide-default-eql-instances/4574
* Third attempt: implement "strict equality" for Scala 2 with a macro
  and `Eql[T, T]` in Scala. This improves the situation for Scala 2,
  but would mean relying on a feature that we can't easily port to Scala 3.
* Fourth attempt (this commit): improve the first attempt (current
  release) by allowing `Compare[A, B]` as long
  as `A <:< B` OR `B <:< A`. This is possible thanks to an observation
  by Gabriele Petronella that it's possible to layer the implicits to
  avoid diverging implicit search.

The benefit of the fourth approach is that it works the same way for
Scala 3 and Scala 3. It's very nice that we can avoid macros as well.

munit/shared/src/main/scala/munit/Compare.scala
```diff
@@ -0,0 +1,115 @@
...
+    assertions.failComparison(
+      s"values are not equal even if they have the same `toString()`: $obtained",
+      obtained,
+      expected
+    )(loc)
+  }
+
+}
+
+object Compare extends ComparePriority1 {
```
olafurpg, line 92: @gabro as discussed offline, we have the option of including these instances by default in the `Compare` companion object, or provide them as an optional mix-in for `FunSuite`. I personally lean towards the companion object because I have a bad experience with type-classes that don't have good defaults in the companion.

gabro: That's ok. We can change this in case we have a use-case where we want to explicitly exclude these instances

munit/shared/src/main/scala/munit/Compare.scala
```diff
@@ -0,0 +1,115 @@
...
+trait ComparePriority1 extends ComparePriority2 {
+  implicit def compareSubtypeWithSupertype[A, B](implicit
+      ev: A <:< B
+  ): Compare[A, B] = Compare.defaultCompare
+}
+
+/**
+ * Allows comparison between A and B when B is a subtype of A.
+ *
+ * This implicit is defined separately from ComparePriority2 in order to avoid
```
gabro, line 108:
```suggestion
 * This implicit is defined separately from ComparePriority1 in order to avoid
```

tests/shared/src/test/scala/munit/AssertionsSuite.scala
```diff
@@ -50,12 +50,132 @@ class AssertionsSuite extends BaseSuite {
...
+           |  Alternative 2: upcast either type into `Any`
+           |assertEquals(new A, new B)
+           |            ^
+           |""".stripMargin
+    )
+  }
+
+  test("chat-int-nok") {
+    assertNoDiff(
+      compileErrors("assertEquals('a', 'a'.toInt)"),
```
gabro, line 143: 🎉

gabro (approved):
Really nice! 💣

olafurpg:
Summarizing the contributors thread in https://contributors.scala-lang.org/t/should-multiversal-equality-provide-default-eql-instances/4574/26?u=olafurpg

The `Compare[A, B]` type-class that's introduced in this PR does not satisfy compile-time transitivity. For example
```
class A
B extends A
C extends A

B == A // OK
A == C // OK
B == C // Error
```
While I agree it's necessary to have transitivity for runtime equality, I am not convinced it's necessary for compile-time equality. The worst-case scenario is that users get a compile-error and need to upcast one of the arguments into the LUB. This is only necessary for classes with custom `override def equals()` semantics, because it's otherwise always an application bug when you compare two sibling subtypes.

valencik:
Closing this as it was revived and merged in https://github.com/scalameta/munit/pull/521

PR #228, opened 2020-10-19
Move `Fixture[T]` into a toplevel class.

### LINKED ISSUE #175: Make Fixture a plain class instead of an inner class of Suite
I would love to be able to create `Fixture` instances outside of a test suite; alas, it is an inner class of `Suite`. 

I currently use munit to create a CLI tool which executes multiple "tests" on a given entity, and that entity is constructed dynamically according to user input:

```scala

// would love this to extend from munit.Fixture
trait MyFixture {
  def createEntity(): Entity

  def beforeAll(): Unit = {}
  // and all the rest
}

abstract class MySuite(fixture: MyFixture) extends munit.Suite {

  protected var entity: Entity = _

  override def beforeAll(): Unit = {
    super.beforeAll()
    fixture.beforeAll()
    entity = createEntity()
  }
}

class MyRunner(cls: Class[_ <: MySuite], newInstance: () => Suite) extends MUnitRunner(suite, newInstance) {
  def this(fixture: MyFixture, cls: Class[_ <: MySuite]) = {
    this(cls, MyUtils.findCtr(cls).newInstance(fixture))
  }
}

object MyApp extends App {
  val input = ???
  val fixture = generateFixture(input)
  val suites = findSuites(input)

  val junit = new JUnitCore
  junit.addListener(new TextListener(System.out))
  suites.foreach { suite =>
    junit.run(new MyRunner(fixture, suite))
  }
}
```

The reason I'm constructing the entity via fixtures while executing the tests (and not outside of the execution) is because its creation is rather complex and requires both resource initialization and mocking (via mockito).

I hope it's a clear enough example, but if not just let me know and I'll further explain :)

olafurpg:
Previously, `Fixture[T]` was declared as an inner class of `FunSuite`.
This made it annoying to implement fixtures as normal toplevel classes.
Now, `Fixture[T]` is just a normal toplevel class making it easier to
reason about.

Breaking change: In order to implement this change, we remove the
ability to extend `Suite` with a custom `type TestValue`. The test value
is now hardcoded to `Future[Any]`. I'm not aware of any library that
extends `munit.Suite` with a custom `TestValue`. All the 3rdparty
integrations that I have seen hardcode against `munit.FunSuite` anyways.
Given this feature doesn't seem to be used anywhere and it complicates
the public API, I feel it's best to remove it.

Fixes #175

Waiting to merge #225 before moving further with this PR.


olafurpg:
Superseded by #430 , which makes `Fixture[T]` a toplevel class.

Issue #240, opened 2020-11-03
Release for Scala 3.0.0-M1


larsrh:
ScalaCheck 1.15.0 has just been deployed to Sonatype, which should allow munit to be released.

tgodzik:
I will take a look at making a PR if no one objects.

larsrh:
A backported release of 0.7.15 would be great, too ☺

tgodzik:
Thaaat might be harder as I have no idea how to release it locally :sweat_smile:

olafurpg:
No need to back port, we can continue publishing new 0.7.x versions.

larsrh:
True, there's no need, but it's a lot more convenient for downstream dependencies. It avoids churn going forward with 3.0.0-M*.

tgodzik:
0.7.16 with support for `0.27.0-RC1` and `3.0.0-M1` is now released.

PR #241, opened 2020-11-03
Update Scala 3 version to 3.0.0-M1

tgodzik:
Connected to https://github.com/scalameta/munit/issues/240

Seems we are no longer getting RC versions, so I reverted back to having single stable Dotty version.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,8 +11,8 @@ def previousVersion = "0.7.0"
 def scala213 = "2.13.2"
 def scala212 = "2.12.11"
 def scala211 = "2.11.12"
-def dottyNext = "0.27.0-RC1"
-def dottyStable = "0.26.0"
+def scala3Stable = "3.0.0-M1"
+def scala3Previous = "0.27.0-RC1"
 def junitVersion = "4.13"
 def gcp = "com.google.cloud" % "google-cloud-storage" % "1.113.2"
 inThisBuild(
@@ -62,11 +62,11 @@ addCommandAlias(
 )
 val isPreScala213 = Set[Option[(Long, Long)]](Some((2, 11)), Some((2, 12)))
 val scala2Versions = List(scala213, scala212, scala211)
-val scala3Versions = List(dottyNext, dottyStable)
+val scala3Versions = List(scala3Previous, scala3Stable)
 val allScalaVersions = scala2Versions ++ scala3Versions
 def isNotScala211(v: Option[(Long, Long)]): Boolean = !v.contains((2, 11))
 def isScala2(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 2)
-def isScala3(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 0)
+def isScala3(v: Option[(Long, Long)]): Boolean = v.exists(_._1 != 2)
 val isScalaJS = Def.setting[Boolean](
   SettingKey[Boolean]("scalaJSUseMainModuleInitializer").?.value.isDefined
 )
@@ -132,7 +132,7 @@ val sharedSettings = List(
           "-Xexperimental",
           "-Ywarn-unused-import"
         )
-      case Some((0, _)) => List()
+      case Some((major, _)) if major != 2 => List()
       case _ =>
         List(
           "-target:jvm-1.8",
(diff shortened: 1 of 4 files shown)


tgodzik:
I guess this will not be that easy after all :(  I will take a further look tomorrow, but it seems that we are now printing full names with packages. If anyone wants to fix it today, you can just push to this PR with a fix.

build.sbt
```diff
@@ -11,8 +11,7 @@ def previousVersion = "0.7.0"
 def scala213 = "2.13.2"
 def scala212 = "2.12.11"
 def scala211 = "2.11.12"
-def dottyNext = "0.27.0-RC1"
-def dottyStable = "0.26.0"
+def dottyStable = "3.0.0-M1"
```
gabro, line 14: how about renaming this to scala3?

olafurpg (approved):
Good to get this in

mpilquist:
Is this related? https://github.com/lampepfl/dotty/issues/9716

tgodzik:
@mpilquist might be, but that issue was also present in the previous versions, so it shouldn't actually change here 🤔

mpilquist:
I just noticed this commit in community build -- trying it now: https://github.com/dotty-staging/munit/commit/1e90e9707e02c41049489849dd0e002d48d4b498

mpilquist:
Alright figured this one out -- nothing changed in Dotty w.r.t. showing type prefixes. The bug is that we need to change `def isDotty` in `BaseSuite` to account for the version number change.

gabro:
Aha, nice catch @mpilquist!

tgodzik:
> Alright figured this one out -- nothing changed in Dotty w.r.t. showing type prefixes. The bug is that we need to change `def isDotty` in `BaseSuite` to account for the version number change.

Thanks! Good catch! I updated the PR and also changed the MacroCompat a bit based on the commit you mentioned before.

build.sbt
tgodzik: Changed it!

tgodzik: And remembered the env variables too in the next push :sweat_smile:

tgodzik:
I will merge and release first thing tomorrow (just in case there is an issue) if no one objects.

mpilquist:
Looks good to me. Once munit publishes, we'll be able to start building the various Typelevel libraries. The sooner the better but if you want to wait until tomorrow, that's great. :)

gabro (approved):
Nice 👏

djspiewak:
> Seems we are no longer getting RC versions, so I reverted back to having single stable Dotty version.

I definitely object. It's very very difficult for downstream projects to juggle incompatibilities as the ecosystem updates. Maintaining builds for the two most recent Dotty releases (whenever possible) makes this significantly easier since it allows libraries to take the upgrade in chunks, particularly if and when munit itself makes changes. Publishing for the latest two dotty releases ensures that users are able to split the upgrade pain into two steps: upgrade munit, then upgrade dotty.

Meanwhile, from the perspective of a library maintainer, *not* dropping the previous Dotty release is usually completely painless. Sometimes it has to be done, and particularly with ScalaJS shenanigans I can see merit to dropping 0.27.0-RC1, but I would strongly recommend that publication be for `Mn` and `M(n-1)` from this point forward. We're about to go through the most convoluted and painful cross-publication coordination effort in Scala's history; the least we can do is try to learn from what has made all of the other (otherwise simpler) migrations difficult at times.

tgodzik:
> > Seems we are no longer getting RC versions, so I reverted back to having single stable Dotty version.
> 
> I definitely object. It's very very difficult for downstream projects to juggle incompatibilities as the ecosystem updates. Maintaining builds for the two most recent Dotty releases (whenever possible) makes this significantly easier since it allows libraries to take the upgrade in chunks, particularly if and when munit itself makes changes. Publishing for the latest two dotty releases ensures that users are able to split the upgrade pain into two steps: upgrade munit, then upgrade dotty.
> 
> Meanwhile, from the perspective of a library maintainer, _not_ dropping the previous Dotty release is usually completely painless. Sometimes it has to be done, and particularly with ScalaJS shenanigans I can see merit to dropping 0.27.0-RC1, but I would strongly recommend that publication be for `Mn` and `M(n-1)` from this point forward. We're about to go through the most convoluted and painful cross-publication coordination effort in Scala's history; the least we can do is try to learn from what has made all of the other (otherwise simpler) migrations difficult at times.

I can add back `0.27.0-RC1` for sure, might have rushed with dropping it.

olafurpg:
@tgodzik feel free to merge and release when this PR is ready!

tgodzik:
@djspiewak I added back the `0.27.0-RC1` version. If everything runs ok, I will merge and release it if no one objects?

Issue #199, opened 2020-09-14
Release Scala.js artifacts for Dotty builds


mpilquist:
As of Dotty 0.27.0-RC1, we can start releasing Dotty Scala.js artifacts.

olafurpg:
Thank you for reporting! I took a stab at implementing this here https://github.com/scalameta/munit/compare/master...olafurpg:dotty-js?expand=1 but I'm getting the error below

```
==> X munit.LinesSuite.basic 0.02s scala.scalajs.js.JavaScriptException: ReferenceError: JSPath is not defined
    at java.io.File$.separator(/Users/lgeirsson/dev/munit/tests/js/target/scala-0.27/tests-test-fastopt.js:733)
    at munit.Location.filename(/Users/lgeirsson/dev/munit/tests/js/target/scala-0.27/tests-test-fastopt.js:27493)
    at <jscode>.test__Lmunit_TestOptions__F0__Lmunit_Location__V.$c_sjsr_AnonFunction0(/Users/lgeirsson/dev/munit/tests/js/target/scala-0.27/tests-test-fastopt.js:53726)
    at scala.scalajs.runtime.AnonFunction0.apply(/Users/lgeirsson/dev/munit/tests/js/target/scala-0.27/tests-test-fastopt.js:32951)
```

If there was a way to make this method here return `false` then I think the problem would be fixed

https://github.com/scalameta/munit/blob/0c4ff57f667970e6574a4c7e02a6296e9feb8990/munit/js/src/main/scala/munit/internal/JSIO.scala#L88

Any help in this area would be greatly appreciated! cc/ @sjrd

sjrd:
The error you are experiencing is a bug in the Scala.js support in dotty, which will be fixed in https://github.com/lampepfl/dotty/pull/9767. You'll probably have to wait for the next release.

Independently, your `def isNode` implementation looks buggy in Scala.js 1.x, whether for Scala 2 or 3. You should use `js.typeOf` instead:
```scala
js.typeOf(process) != "undefined" && js.typeOf(process.cwd) != "undefined"
```
The `js.isUndefined(process)` will throw a `ReferenceError` if `process` is not defined. (In 0.6.x it would have returned `undefined`, but not in 1.x.)

olafurpg:
Thank you @sjrd ! I will wait until the next release then.

smarter:
> I will wait until the next release then.

[...]

PR #243, opened 2020-11-04
Enable Scala.js builds on 3.0.0-M1

mpilquist:


olafurpg:
Thank you for this contribution! Does Scala 3.0.0-M1 support JS exports? Judging by https://github.com/scalameta/munit/issues/199#issuecomment-692705511 and https://scala-lang.org/2020/11/03/scalajs-for-scala-3.html I thought this part was still blocking Scala 3 JS support for MUnit

> The only missing feature left is JS exports, which we will implement by the next milestone.

olafurpg:
I suspect most of MUnit's functionality will work without JS exports, but it would require changes to the build to make the CI green. If JS exports will become supported in the next milestone release then I think it's worth waiting

mpilquist:
@olafurpg Scalacheck SJS build is also blocked by some erasure issue, so maybe it's best to punt SJS support to 3.0.0-M2

PR #248, opened 2020-11-08
enable DottyJS, drop Scala.js 0.6.x

### LINKED ISSUE #199: Release Scala.js artifacts for Dotty builds

larsrh:
Drop Scala.js 0.6.x and enable DottyJS for 3.0.0-M1.

I haven't enabled DottyJS for 0.27 because tests are somehow failing there.

Should help fix #199. Obsoletes #246.

### DIFF
--- a/munit/non-jvm/src/main/scala/munit/internal/junitinterface/Settings.scala
+++ b/munit/non-jvm/src/main/scala/munit/internal/junitinterface/Settings.scala
@@ -1,11 +1,11 @@
 package munit.internal.junitinterface
 
 trait Settings {
-  def trimStackTraces(): Boolean
+  def trimStackTraces: Boolean
 }
 
 object Settings {
   def defaults(): Settings = new Settings {
-    def trimStackTraces(): Boolean = true
+    def trimStackTraces: Boolean = true
   }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,15 +23,12 @@ jobs:
   js:
     strategy:
       fail-fast: false
-      matrix:
-        scalajs: ["1.3.0", "0.6.32"]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: olafurpg/setup-scala@v7
       - run: sbt +testsJS/test
         env:
-          SCALAJS_VERSION: ${{ matrix.scalajs }}
           GOOGLE_APPLICATION_CREDENTIALS:
             ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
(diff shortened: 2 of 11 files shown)


munit/non-jvm/src/main/scala/munit/internal/junitinterface/Settings.scala
```diff
@@ -1,11 +1,11 @@
 package munit.internal.junitinterface
 
 trait Settings {
-  def trimStackTraces(): Boolean
```
olafurpg, line 4: Can we add the parentheses back here? These interfaces are meant to be mirrors of their equivalent Java interfaces https://github.com/scalameta/munit/blob/7ea3075ae51ca73c7fa84373aa9dd82cf9a6ccd7/junit-interface/src/main/java/munit/internal/junitinterface/Tag.java

I prefer to keep the two synchronized as much as possible, these are internal APIs so it shouldn't matter much whichever we use.

olafurpg (changes requested):
Thank you for this contribution!

I'm OK with dropping 0.6.x support. Can we update the 0.6.x column on the website to list the 0.7.16 as the last version supporting 0.6.x? https://scalameta.org/munit/docs/getting-started.html

Screenshot 2020-11-08 at 12 19 44

A few nitpick comments regarding parentheses, otherwise looks good to me!

munit/non-jvm/src/main/scala/munit/internal/junitinterface/Settings.scala
larsrh: I can add them back, but that makes stuff in some other places more complicated. There are some classes that attempt to use `val` to override those `def`s (which doesn't work anymore in Dotty, so they have to be renamed). Those constructors are subsequently called with named parameters, so their call sites have to be renamed too.

I'm happy to go ahead either way, though.

olafurpg: I see. That makes sense, I'm glad Scala 3 is more strict about the differences between `method()` and `method` without parentheses. It's still quite a large breaking change, I always thought it was elegant that `val` could override parameterless defs. 

Anyways, I'm fine dropping the parentheses in this case 👍

olafurpg (changes requested):
No need to change the parentheses, just the Scala.js 0.6 support mention in the docs.

olafurpg (approved):
LGTM 👍 Thank you!

olafurpg:
Kickstarted 0.7.17 https://github.com/scalameta/munit/actions/runs/354125829

PR #256, opened 2020-11-17
Fix the CI and configure dependabot to keep actions updated

gabro:
This PR updates the setup-scala action to v10, which fixes the CI error seen on https://github.com/scalameta/munit/pull/255/checks

It also adds a dependabot configuration that will keep actions up-to-date


tgodzik (approved):
LGTM! I also updated the scalameta repo

PR #264, opened 2020-11-25
Cross-build against latest Scala 3 milestone release.

olafurpg:
Fixes #263

PR #265, opened 2020-11-25
Cross-build against Scala 3.0.0-M2

### LINKED ISSUE #263: 3.0.0-M2 release
Scalacheck was just published so munit should be ready.

gabro:
Supersedes #264 

Closes #263


munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala
```diff
@@ -10,21 +10,22 @@ object MacroCompat {
     inline implicit def generate: Location = ${ locationImpl() }
   }
 
-  def locationImpl()(using qctx: QuoteContext): Expr[Location] = {
-    import qctx.tasty.{_, given _}
-    val path = rootPosition.sourceFile.jpath.toString
-    val startLine = rootPosition.startLine + 1
+  def locationImpl()(using Quotes): Expr[Location] = {
```
tgodzik, line 13: Looks like Quotes are not found, I also wonder if this will compile under previous versions? 3.0.0-M1 and 0.27.0-RC1?

I think we can drop 0.27.0-RC1, but we might need to have some campatibility utility classes after all.

gabro: I was stupid and I didn't consider M1. Should be fixed in the latest commit

gabro: I don't think we should drop 0.27-RC1 in this PR though, maybe we can discuss the consequences in a different issues

gabro:
Ok, since the macro API seems to be evolving across dotty and scala milestones, I've opted for splitting `MacroCompat` across the specific Scala versions (0.27, 3.0.0-M1, 3.0.0-M2, etc) and leaving `Compat` in the shared `scala-3` dir.

This is with the idea that once Scala 3 is stable we can get rid of most of them.

olafurpg (approved):
Thank you for fixing this! Feel free to cut a new release once merged

gabro:
👍 I'll merge and cut a release and then we can discuss dropping 0.27

PR #270, opened 2020-11-30
Adding minimal setup notes for Mill and Scala.js

davesmith00000:
Hi,

I've added a few notes that I would have found very useful and would have saved me an hour or two of trying to make it work.

I am in no way precious about the wording or layout, please feel free to reject and rephrase - it's quite late here! :-)

All the best,

Dave


olafurpg (approved):
LGTM 👍 This is great! Thank you 🙏

PR #277, opened 2020-12-14
Unwrap "Boxed Error" exceptions, fixes #275

### LINKED ISSUE #275: assert fail throws Boxed Error
When this code is executed I get a Boxed Error, which can be confusing
```scala
class Test extends munit.FunSuite {

  case class Inner(value: Double)

  case class Outer(data: Seq[Inner])

  test("exist issue") {
    val values = Array(
      Outer(Seq(Inner(1), Inner(2), Inner(10))),
      Outer(Seq(Inner(1), Inner(2), Inner(10))),
      Outer(Seq(Inner(1), Inner(2), Inner(10)))
    )
    assert(values.exists(outer => outer.data.exists(inner => inner.value > 90)))
  }

}
```
Result:
```
java.util.concurrent.ExecutionException: Boxed Error

    at scala.concurrent.impl.Promise$.resolver(Promise.scala:87)
    at scala.concurrent.impl.Promise$.scala$concurrent$impl$Promise$$resolveTry(Promise.scala:79)
    at scala.concurrent.impl.Promise$KeptPromise$.apply(Promise.scala:406)
    at scala.concurrent.Promise$.fromTry(Promise.scala:142)
    at scala.concurrent.Future$.fromTry(Future.scala:640)
    at munit.ValueTransforms.munitValueTransform(ValueTransforms.scala:37)
    at munit.ValueTransforms.munitValueTransform$(ValueTransforms.scala:22)
    at munit.FunSuite.munitValueTransform(FunSuite.scala:11)
    at munit.FunSuite.$anonfun$test$1(FunSuite.scala:37)
    at munit.MUnitRunner.$anonfun$runTestBody$1(MUnitRunner.scala:290)
Caused by: munit.FailException: /home/ale/workspace/munit_issue/src/test/scala/Test.scala:13 assertion failed
12:    )
13:    assert(values.exists(outer => outer.data.exists(inner => inner.value > 90)))
14:  }
    at munit.FunSuite.assert(FunSuite.scala:11)
    at Test.$anonfun$new$1(Test.scala:13)
    at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
```
Expected:
```
munit.FailException: /home/ale/workspace/munit_issue/src/test/scala/Test.scala:13 assertion failed
12:    )
13:    assert(values.exists(outer => outer.data.exists(inner => inner.value > 90)))
14:  }

    at munit.FunSuite.assert(FunSuite.scala:11)
    at Test.$anonfun$new$1(Test.scala:13)
    at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
```
If the assertion return true everything works as expected
munit version: 0.7.19
scala version: 2.11.12, 2.12.12

scala version 2.13.4 works as expected

olafurpg:
This makes the error message clearer in some cases.

### DIFF
--- a/munit/shared/src/main/scala/munit/MUnitRunner.scala
+++ b/munit/shared/src/main/scala/munit/MUnitRunner.scala
@@ -4,7 +4,9 @@ import munit.internal.junitinterface.Configurable
 import munit.internal.PlatformCompat
 import org.junit.runner.Description
 import org.junit.runner.notification.RunNotifier
+import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Modifier
+import java.lang.reflect.UndeclaredThrowableException
 
 import munit.internal.FutureCompat._
 import munit.internal.console.StackTraces
@@ -245,19 +247,7 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
         Future.successful(())
       case NonFatal(ex) =>
         trimStackTrace(ex)
-        val cause = ex match {
-          case e: ExecutionException if e.getCause() != null && {
-                e.getMessage() match {
-                  // NOTE(olafur): these exceptions appear when we await on
-                  // futures. We unwrap these exception in order to provide more
-                  // helpful error messages.
-                  case "Boxed Exception" | "Boxed Error" => true
-                  case _                                 => false
-                }
-              } =>
-            e.getCause()
-          case e => e
-        }
+        val cause = rootCause(ex)
         val failure = new Failure(description, cause)
         cause match {
           case _: AssumptionViolatedException =>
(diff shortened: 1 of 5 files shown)


munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -246,9 +246,15 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
-          case e: ExecutionException
-              if "Boxed Exception" == e.getMessage() &&
-                e.getCause() != null =>
+          case e: ExecutionException if e.getCause() != null && {
+                e.getMessage() match {
+                  // NOTE(olafur): these exceptions appear when we await on
+                  // futures. We unwrap these exception in order to provide more
+                  // helpful error messages.
+                  case "Boxed Exception" | "Boxed Error" => true
+                  case _                                 => false
```
gabro, line 255: do we need the check on `getMessage`? I think something like

https://github.com/scala/scala/blob/ba9701059216c629410f4f23a2175d20ad62484b/src/compiler/scala/tools/nsc/util/Exceptional.scala#L19-L27

would work here (and it's nice to handle also other cases)

gabro (commented):
aha, thanks @olafurpg I didn't think about doing this in `MUnitRunner`

munit/shared/src/main/scala/munit/MUnitRunner.scala
olafurpg: Nice! That ends up being way more general, the old solution was not nice at all

olafurpg:
Thank you @gabro! That was a great suggestion 🙏

gabro (approved):
Nice!

PR #279, opened 2020-12-17
Add support for Scala 3.0.0-M3

tgodzik:
It seems that scalacheck was wrongly released for 1.5.1, which is not an existing version, so I needed to go around it for now. The issue is tracked under https://github.com/typelevel/scalacheck/issues/732

I also unignored some tests thanks to recent changes for show methods and dropped 0.27.0-RC1


munit/shared/src/main/scala-3.0.0-M3/munit/internal/MacroCompat.scala
```diff
@@ -10,21 +10,22 @@ object MacroCompat {
...
   }
 
-  def clueImpl[T:Type](value: Expr[T])(using qctx: QuoteContext): Expr[Clue[T]] = {
-    import qctx.tasty.{_, given _}
-    val source = value.unseal.pos.sourceCode
-    val valueType = implicitly[scala.quoted.Type[T]].show
+  def clueImpl[T: Type](value: Expr[T])(using Quotes): Expr[Clue[T]] = {
+    import quotes.reflect._
+    val source = Term.of(value).pos.sourceCode.getOrElse("")
+    val valueType = TypeTree.of[T].show(using Printer.TreeShortCode)
```
tgodzik, line 28: This will print type without the package. This and one line above is the only real change for 3.0.0-M3

build.sbt
```diff
@@ -251,7 +251,13 @@ lazy val munitScalacheck = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(
     moduleName := "munit-scalacheck",
     sharedSettings,
-    libraryDependencies += "org.scalacheck" %%% "scalacheck" % "1.15.1"
+    libraryDependencies += {
+      // scalcheck was wrongly released for 1.5.1 instead of 1.15.1
+      if (scalaVersion.value == "3.0.0-M3")
+        "org.scalacheck" %%% "scalacheck" % "1.5.1"
```
larsrh, line 257: I highly recommend against doing that, this will create too much pain downstream

gabro (approved):
LGTM, apart from the Scalacheck issue for which I'd wait for the new release

Thanks @tgodzik!

larsrh:
1.15.2 is promoting to Maven Central right now

build.sbt
```diff
@@ -10,8 +10,8 @@ def previousVersion = "0.7.0"
 def scala213 = "2.13.2"
 def scala212 = "2.12.11"
 def scala211 = "2.11.12"
-def scala3Stable = "3.0.0-M2"
-def scala3Previous = List("3.0.0-M1", "0.27.0-RC1")
+def scala3Stable = "3.0.0-M3"
+def scala3Previous = List("3.0.0-M2", "3.0.0-M1")
```
mpilquist, line 14: The new ScalaCheck release dropped support for 3.0.0-M1 so we'll need to drop here too (or at least that's the simplest option)

gabro: I agree. Unless we have a very good reason, keeping only the two latest versions of Scala 3 around should be enough for smooth downstream support

build.sbt
tgodzik: Fixed, thanks for the help!

build.sbt
tgodzik: Removed it and left `3.0.0-M2` and `3.0.0-M3` only.

PR #283, opened 2021-01-05
Mix Scala 2 macros in MUnit Scala 3 module

olafurpg:
With this change, Scala 2.13 users will be able to depend on munit_3.0
artifacts instead of munit_2.13.

I wasn't able to manually test this change because MUnit doesn't
cross-build for 3.0.0-M1, which is the latest version supported by the
`-Ytasty-reader` flag in 2.13.4. Once 2.13.5 is out, it will be easier
ot make use of this feature.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -127,7 +127,10 @@ val sharedSettings = List(
           "-Xexperimental",
           "-Ywarn-unused-import"
         )
-      case Some((major, _)) if major != 2 => List()
+      case Some((major, _)) if major != 2 =>
+        List(
+          "-language:implicitConversions"
+        )
       case _ =>
         List(
           "-target:jvm-1.8",
--- a/build.sbt
+++ b/build.sbt
@@ -40,11 +40,7 @@ inThisBuild(
     testFrameworks := List(
       new TestFramework("munit.Framework")
     ),
-    useSuperShell := false,
-    scalacOptions ++= List(
-      "-target:jvm-1.8",
-      "-language:implicitConversions"
-    )
+    useSuperShell := false
   )
 )
 
@@ -187,15 +183,12 @@ lazy val munit = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       }
       result.toList
     },
-    libraryDependencies ++= {
-      CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((major, _)) if major != 2 => Nil
-        case _ =>
-          List(
-            "org.scala-lang" % "scala-reflect" % scalaVersion.value
-          )
-      }
-    }
+    libraryDependencies ++= List(
+      "org.scala-lang" % "scala-reflect" % {
+        if (isDotty.value) scala213
+        else scalaVersion.value
+      } % Provided
+    )
   )
   .nativeConfigure(sharedNativeConfigure)
   .nativeSettings(
--- a/build.sbt
+++ b/build.sbt
@@ -125,7 +125,8 @@ val sharedSettings = List(
         List(
           "-Yrangepos",
           "-Xexperimental",
-          "-Ywarn-unused-import"
+          "-Ywarn-unused-import",
+          "-target:jvm-1.8"
         )
       case Some((major, _)) if major != 2 =>
         List(
--- a/build.sbt
+++ b/build.sbt
@@ -61,6 +61,10 @@ val scala3Versions = scala3Stable :: scala3Previous
 val allScalaVersions = scala2Versions ++ scala3Versions
 def isNotScala211(v: Option[(Long, Long)]): Boolean = !v.contains((2, 11))
 def isScala2(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 2)
+val isScala3Setting = Def.setting {
+  isScala3(CrossVersion.partialVersion(scalaVersion.value))
+}
+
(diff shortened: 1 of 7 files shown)


website/blog/2021-01-05-macromix.md
```diff
@@ -0,0 +1,184 @@
+---
+author: Ólafur Páll Geirsson
+title: Publish Scala 2 and Scala 3 macros together
+authorURL: https://twitter.com/olafurpg
+authorImageURL: https://github.com/olafurpg.png
+---
+
+The next release of MUnit makes use of a new compiler feature that allows you to
```
olafurpg, line 8: cc/ @bishabosha @sjrd could you please take a look at this blog post? Please correct me if I got the details wrong

olafurpg: Live preview at https://olafurpg.github.io/munit/blog/2021/01/05/macromix.html so you can see the rendered SVG files.

gabro (commented):
Really cool!

What's the longer plan though? Do we still publish MUnit also for 2.13? If not, I assume Scala 2 users will need to use the full artifact name in sbt to depend on the 3.0 artifact? Either way it may be good to add some notes about this in the user documentation, what do you think?

build.sbt
```diff
@@ -187,15 +183,12 @@ lazy val munit = crossProject(JSPlatform, JVMPlatform, NativePlatform)
...
-        case Some((major, _)) if major != 2 => Nil
-        case _ =>
-          List(
-            "org.scala-lang" % "scala-reflect" % scalaVersion.value
-          )
-      }
-    }
+    libraryDependencies ++= List(
+      "org.scala-lang" % "scala-reflect" % {
+        if (isDotty.value) scala213
```
gabro, line 188: Does isDotty come from the sbt-dotty plug-in? Maybe we could use isScala3 for consistency with the rest of the build

website/blog/2021-01-05-macromix.md
```diff
@@ -0,0 +1,183 @@
...
+2.13.4 compiler that enables the Scala 2 compiler to read Scala 3 libraries.
+
+This feature is new and has some limitations that's good to be aware of:
+
+- Scala 2.13.4 can only understand libraries that are published with the old
+  Scala 3.0.0-M1 version (latest is 3.0.0-M3 at the time of this writing). The
+  upcoming Scala 2.13.5 release will be able to understand newer Scala 3.x
+  releases.
+- Some common features in Scala 2 macros don't work in Scala 3. Most notably,
+  you can't use quasiquotes or macro bundles. In the case of MUnit, we had to
```
bishabosha, line 175: I was confused to find that I could compile a macro bundle using Scala 3.0.0-M1 but the macro expansion failed from Scala 2.13.4, despite the macro bundle having the same bytecode when compiled by either 2.13.4 or 3.0.0-M1, and the tasty to call the macro was the same either way. Something to do with reflective invocation not having the required number of arguments.

It's probably worth opening an issue with the trouble you found, because I don't see at this moment why it should fail.

my example was the following:
```scala
class TpeTagImpl(val c: Context) {
  import c.universe._

  // factory for `final case class TpeTag[T](repr: String)`
  // invoked by `implicit def mkTpeTag[T]: TpeTag[T] = macro TpeTagImpl.mkTag[T]`
  def mkTag[T](implicit T: c.WeakTypeTag[T]): Tree = {
    val tpeString = Literal(Constant(T.tpe.toString))
    // new testframework.TpeTag[T](T.tpe.toString)
    New(appliedType(c.mirror.staticClass(classOf[testframework.TpeTag[_]].getName()).toType, T.tpe), tpeString)
  }

}
```

build.sbt
olafurpg: Good point. Fixed.

olafurpg:
@gabro thank you for the review!

> What's the longer plan though? Do we still publish MUnit also for 2.13? If not, I assume Scala 2 users will need to use the full artifact name in sbt to depend on the 3.0 artifact? Either way it may be good to add some notes about this in the user documentation, what do you think?

I was thinking of keeping things unchanged. I updated the post to explain that MUnit 2.13 will continue to be published.

The primary purpose of this post is to advocate for this route since I don't everybody are fully aware of the risks of cross-building 2.13 and 3.

bishabosha (approved):

website/blog/2021-01-05-macromix.md
olafurpg: I removed the macro bundle mention. I didn't try using bundles, I just assumed they wouldn't work because I thought they were implemented with macroparadise.

olafurpg: I confused bundles with https://github.com/milessabin/macro-compat, which is implemented as a macro annotation to add bundle support in Scala 2.11

sjrd (approved):
A bit late, but FWIW the blog post LGTM.

olafurpg:
@sjrd whoops 😅 I should have waited for your LGTM before merging.

website/blog/2021-01-05-macromix.md
bishabosha: This is very nice work, I will open the issue then for my example.

sjrd:
No worries. I could have looked at this earlier this morning. ;)

Issue #288, opened 2021-01-12
Offer a trait version of FunSuite


vincenzobaz:
`FunSuite` is a class. It would be nice to also have a `trait FunSuiteBase` so that test suites can extend another class and mix in all the testing equipment.

PR #289, opened 2021-01-12
Expose a FunSuite trait while preserving API

vincenzobaz:
This PR addresses issue #288


julienrf:
@vincenzobaz would it be possible to keep binary compatibility by having the same constructor signatures, or maybe by adding them back as secondary constructors?

https://github.com/scalameta/munit/pull/289/checks?check_run_id=1959596809#step:4:249

olafurpg:
Thank you for this contribution! I opened #433 which implements this change on top of the latest master and uses the name `BaseFunSuite` instead of `FunSuiteBase` to prevent the situation where IDEs will auto-complete `FunSuiteBase` as you're typing `FunSuite`. This is a binary breaking change but that's OK since we're preparing a v1.0.0 release, which includes other binary breaking change like #418

PR #297, opened 2021-01-19
Update Scala versions to 2.12.13 and 2.13.4

tgodzik:
It seems that the release is failing due to the changes in mdoc I did where we now release:
`mdoc_2.11`, `mdoc_2.12.12`, `mdoc_2.12` and `mdoc_2.13`, where `mdoc_2.12.12` was added to support Scala versions prior to 2.12.13 due to binary incompatibility.

Somehow when using Scala 2.12.11 it actually needs `mdoc-runtime_2.12.11`, which seems to be a bug in coursier. I reported it here: https://github.com/sbt/sbt/issues/6269

This fixes the issue, but I need to investigate a bit further and will try to work around it in mdoc.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,8 +7,8 @@ import scala.collection.mutable
 val scalaJSVersion = "1.4.0"
 val scalaNativeVersion = "0.4.0-M2"
 def previousVersion = "0.7.0"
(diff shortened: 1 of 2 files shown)


gabro (approved):
Thanks!

PR #299, opened 2021-01-20
Use 'main' instead of 'master' in GitHub Actions workflows

olafurpg:


gabro (approved):
ah, I knew I had missed something! Thanks 🙏

PR #302, opened 2021-01-27
Add more details on how to setup test reports on google cloud

tgodzik:


gabro (approved):
LGTM!

tgodzik:
btw. the CI is also failing here: https://github.com/scalameta/munit/runs/1783178213?check_suite_focus=true#step:4:315

We should either remove it or maybe set it all up under the scalameta account? Don't want to (and most likely I shouldn't) point everything to my account :s 

I wonder what were the monthly cost, if it was not too much, maybe we should just setup some billing info? Or maybe we could somehow keep it all on github?

PR #305, opened 2021-02-04
Remove documentation on how to setup test reports.

olafurpg:
The test reports feature stopped working after my Google Cloud trial
expired. I never used the test reports so it's easier to remove them
instead of updating the GCP credentials.


olafurpg:
Merged this into main.

gabro:
Can I ask why? The instructions were valid anyway, no? 

(I'm asking since I think the feature is pretty cool and I was gonna talk about it in my upcoming talk :D )

tgodzik:
We should be able to just use the reports with different credentials, I managed to set it up in Metals, so let's at least leave the docs. We can remove the reports from munit if they are unused.

olafurpg:
I’m sorry I rushed too quickly to merge this. I’m ok with adding the page back. My impression was only that this feature is not used much. I generally want to avoid documenting functionality that’s half complete.

Issue #311, opened 2021-02-16
Release for Scala 3.0.0-RC1, please?


SethTisue:

tgodzik:
https://github.com/typelevel/scalacheck/pull/759 is merged so we should be soon able to release. Thanks for the reminder!

larsrh:
There's currently a problem:

```
[error] (jvm / doc / scalaInstance) Couldn't retrieve `org.scala-lang:scala3doc:3.0.0-RC1`.
[error] Total time: 33 s, completed Feb 16, 2021, 4:51:19 PM
sbt '+ jvm/publishSigned' failed: 1
```

Only nightlies are available here:

https://repo1.maven.org/maven2/org/scala-lang/scala3doc_3.0.0-RC1/

gabro:
we can easily bypass `doc` for publishing, but it would be nicer to have all the dependencies lined up

gabro:
@tgodzik I can take care of it if you didn't start already

tgodzik:
I added a fix to make it compile in https://github.com/scalameta/munit/pull/312 but not sure about what you mean exactly about bypassing @gabro

tgodzik:
@larsrh I think the name changed? There is https://repo1.maven.org/maven2/org/scala-lang/scaladoc_3.0.0-RC1/3.0.0-RC1/

gabro:
@tgodzik I just meant skipping the `doc` task on Scala 3 to move on with the publishin

tgodzik:
@gabro We can do that for sure, but it seems to work with newest dotty plugin and sbt versions.

larsrh:
> I think the name changed?

Maybe? But I think this requires an update in sbt-dotty. I don't think I can configure the name in the build.

larsrh:
lampepfl/dotty#11416

tgodzik:
> > I think the name changed?
> 
> Maybe? But I think this requires an update in sbt-dotty. I don't think I can configure the name in the build.

Actually it doesn't work with the newest, so you are right. Let me raise it somewhere.

larsrh:
Scalacheck release started

larsrh:
1.15.3 out

SethTisue:
did it actually publish, or did this ticket just auto-close?

tgodzik:
I can release a new version unless anyone else is working on it yet?

tgodzik:
Running it here https://github.com/scalameta/munit/actions/runs/572912654

[...]

PR #312, opened 2021-02-16
Update to Scala 3.0.0-RC1

tgodzik:
Fixes https://github.com/scalameta/munit/issues/311


munit/shared/src/main/scala-3.0.0-RC1/munit/internal/MacroCompat.scala
```diff
@@ -0,0 +1,52 @@
+package munit.internal
```
gabro, line 1: what was the reason for this compat file? (out of curiosity)

gabro: Also, we can probably remove the equivalent M2 file

tgodzik: One small change `val source = value.asTerm.pos.sourceCode.getOrElse("")`

Maybe that is already thee in M3 ?

gabro: it seems worth trying

tgodzik: Ok, it seems to compile ok. Though I am still following up about the docs.

larsrh:
Created a PR for sbt updates: #313

gabro:
Now that #313 is in, rebasing this should fix all the outstanding issues 🤞

gabro (commented):
We also need to update the compatibility table in the docs

tgodzik:
> We also need to update the compatibility table in the docs

I think that's updated automatically.

build.sbt
```diff
@@ -250,7 +250,13 @@ lazy val munitScalacheck = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(
     moduleName := "munit-scalacheck",
     sharedSettings,
-    libraryDependencies += "org.scalacheck" %%% "scalacheck" % "1.15.2"
+    libraryDependencies += {
+      val partialVersion = CrossVersion.partialVersion(scalaVersion.value)
+      if (isNotScala211(partialVersion))
+        "org.scalacheck" %%% "scalacheck" % "1.15.3"
+      else
+        "org.scalacheck" %%% "scalacheck" % "1.15.2"
```
gabro, line 258: @tgodzik / @olafurpg I wonder when will it be a good time to start dropping 2.11 🤔

gabro:
> I think that's updated automatically.

@tgodzik you're right, sometimes I forgets this nice treats of automation 🍬

build.sbt
olafurpg: If ScalaCheck has dropped 2.11 then I won’t oppose dropping it.

olafurpg: We use 2.11 in Metals so maybe it’s worth supporting for that usecase alone.

gabro: Right. I would then probably defer it to when Metals will drop 2.11 (it seems fair to do it in a bit, since native is now supporting more recent Scala versions)

gabro:
@tgodzik found this reference to M2:

https://github.com/scalameta/munit/commit/8034155c14e3c90e9360a23289ca9eff2cc58298

We can probably remove the `OnlyFromM3` tag altogether.

Plus we need to update the test-release script

tgodzik:
Thanks @gabro! I think it should be finally ready :sweat_smile:

gabro (approved):
Looking great! :shipit:

build.sbt
olafurpg: I think 2.11 is still quite heavily used by companies on spark. I have mixed feelings about having deprecated 2.11 in Metals. I haven’t looked at the download numbers to confirm.

gabro: Yeah, let's look at the numbers and see what's the situation. On one end, those numbers are indicative, but on the other they're not independent: if the ecosystem nudges you to move forward, you probably do, at some point. And Metals is part of the ecosystem.

PR #313, opened 2021-02-16
update sbt versions

larsrh:
Subsumes #304 and #301


larsrh:
The fact that the tests are failing here is very unsettling :cold_sweat:

SethTisue:
Partial investigation:
* the problem is reproducible locally, it's not some CI-only glitch
* it's the sbt version bump, not the sbt-dotty bump
* 2.13 passes, 2.11 and 2.12 fail

SethTisue:
oh, I found a ticket on this: https://github.com/sbt/sbt/issues/5934

gabro:
I admit I've never bothered to upgrade Sbt because I didn't want to get into that issue 😅

Wild guess: something related to escaping of output? It seems that the value names are not being printed

mzuehlke:
Updates didn't worked since sbt version 1.4.0.
see #213

SethTisue:
> something related to escaping of output? It seems that the value names are not being printed

It's not a printing problem. In `MacroCompatScala2.scala`, in `clueImpl`, if I `println(text)` after `val text = ...`, it's printing empty strings (on 2.11 and 2.12 only). So the `Clue` values themselves incorrectly contain empty strings.

gabro:
Wow. The fact that this has something to do with sbt is deeply unnerving 😳

I'll take a look when I have a minute

gabro:
@larsrh Nice find, that would explain the differences between 2.11/2.12 and 2.13

mpilquist:
It's definitely the `-Yrangepos` setting. On current head, we see it on testsJVM project when running `+show Test/scalacOptions`. Doing same with this PR shows `-Yrangepos` missing. Not sure what changed upstream to cause this but we could patch easily by setting `Test / scalacOptions := (Compile / scalacOptions).value`

gabro:
I can confirm the workaround proposed by @mpilquist works locally.

build.sbt
```diff
@@ -138,7 +138,8 @@ val sharedSettings = List(
           "-Ywarn-unused:imports"
         )
     }
-  }
+  },
+  Test / scalacOptions := (Compile / scalacOptions).value
```
gabro, line 142: Can we also add a comment pointing to https://github.com/sbt/sbt/issues/5934, so that it's clear why we need this?

gabro (commented):
Code LGTM, just a minor comment 👍

gabro (approved):
🙏 thank you @larsrh and everyone who contributed to this distributed bug hunt!

PR #326, opened 2021-03-08
avoid re-calculating s.length

robstoll:


SethTisue:
(I doubt it matters performance-wise, but I like it because the new code is slightly clearer.)

gabro (approved):
Looks goody thanks!

PR #333, opened 2021-03-23
Fix include/exclude categories arguments in docs

tpetrucciani:
I've changed the include/exclude categories example to be `*-categories=` instead of `*-category=`, to match https://github.com/scalameta/munit/blob/main/junit-interface/src/main/java/munit/internal/junitinterface/JUnitRunner.java#L68.

### DIFF
--- a/docs/filtering.md
+++ b/docs/filtering.md
@@ -211,15 +211,15 @@ class MyFastSuite extends munit.FunSuite {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
LGTM 👍 Thank you!

Reminds me that the website is sorely missing a table of reference for all of the supported command-line flags 😅

PR #344, opened 2021-04-05
Upgrade to sbt 1.5.0

gabro:
Summary of the main changes (I've split them by commit to ease the review):

- Remove sbt-dotty in favor of native support (it requires replacing `withDottyCompat`)
- Remove workaround for an sbt bug around semanticdb scalac options
- Remove deprecated sbt syntax in favor of the new unified slash syntax
- Switch from csbt to sbt on Windows CI (to fix a classloading issue)
- Remove `MUnitReporterPluin` from our docs (sort of a followup of #305), since it was blocking the CI due to invalid credentials
- Update the getting started docs to add MUnit as a custom sbt test framework only on < 1.5.0


SethTisue:
🙀 

```
[error] com.google.cloud.storage.StorageException: The project to be billed is associated with a closed billing account.
[error] 	at com.google.cloud.storage.spi.v1.HttpStorageRpc.translate(HttpStorageRpc.java:233)
```

gabro:
@SethTisue Yeah, that's a known issue I've been avoiding to deal with. The account was @olafurpg's personal account and it expired IIRC. I will disable it for the time being.

gabro:
✅ 🎉

olafurpg (approved):
LGTM 👍 Thank you!

@SethTisue I think it's a cool idea to collect historical test reports but the MUnit implementation wasn't fully fleshed out. For example, there was no good way to properly query/filter the data if you only want a report for the last month or something like that.

PR #359, opened 2021-05-11
Scala 3.0.0

gabro:
It's not out yet, but since it may be this week, I figured we could use some head start 😅

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,8 +10,8 @@ def previousVersion = "0.7.0"
 def scala213 = "2.13.4"
 def scala212 = "2.12.13"
 def scala211 = "2.11.12"
-def scala3Stable = "3.0.0-RC3"
-def scala3Previous = List("3.0.0-RC2")
+def scala3Stable = "3.0.0"
+def scala3Previous = List("3.0.0-RC3")
 def junitVersion = "4.13.2"
 def gcp = "com.google.cloud" % "google-cloud-storage" % "1.113.16"
 inThisBuild(
@@ -305,10 +305,7 @@ lazy val docs = project
       (ThisBuild / baseDirectory).value / "website" / "target" / "docs",
     mdocExtraArguments := List("--no-link-hygiene"),
     mdocVariables := Map(
-      "VERSION" -> version.value.replaceFirst("\\+.*", ""),
-      "SCALA3_PREVIOUS_VERSION" -> scala3Stable,
-      "SCALA3_STABLE_VERSION" -> scala3Previous.head,
-      "SUPPORTED_SCALA_VERSIONS" -> allScalaVersions.mkString(", ")
+      "VERSION" -> version.value.replaceFirst("\\+.*", "")
     ),
     fork := false
   )
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -69,13 +69,12 @@ object test extends Tests {
 }
 ```
 
-| Scala Version             | JVM | Scala.js (0.6.x) | Scala.js (1.x) | Native (0.4.x) |
-| ------------------------- | :-: | :--------------: | :------------: | :------------: |
-| 2.11.x                    | ✅  | ✅ until 0.7.16  |       ✅       |       ✅       |
-| 2.12.x                    | ✅  | ✅ until 0.7.16  |       ✅       |       ✅       |
-| 2.13.x                    | ✅  | ✅ until 0.7.16  |       ✅       |       ✅       |
-| @SCALA3_PREVIOUS_VERSION@ | ✅  |       n/a        |       ✅       |      n/a       |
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -10,8 +10,7 @@ def previousVersion = "0.7.0"
 def scala213 = "2.13.4"
 def scala212 = "2.12.13"
 def scala211 = "2.11.12"
-def scala3Stable = "3.0.0-RC3"
```
tgodzik, line 13: Let's leave RC3 for a while if possible.

gabro: One thing we could do is to drop the support, but backpublish 3.0.0 for 0.7.25. So we the next release would drop RC3, but in the meantime we can keep using 0.7.25 for both 3.0.0 and 3.0.0-RC3

tgodzik: We can work around it in Metals for sure, just last time we dropped to one version there was a bit of a backlash, since people would need to update munit and Scala version at the same time.

gabro: Last time I think @djspiewak had some opinion about this. Maybe also @larsh can chime in?

olafurpg: I think it makes sense to support the old and new version for at least one release. It simplifies upgrading in many cases.

gabro: @olafurpg wdyt about backpublishing 3.0.0 for the current version and merging the changes as is? This way we can move forward in a clean way (scalacheck recently dropped support for 2.11 in a similar fashion)

olafurpg: That sounds fine. I generally avoid backpublishing since it requires manual local steps that’s easy to mess up compared to automated CI releases. It’s a bummer because backpublishing is very user friendly.

larsrh: 1. Yes, please keep two versions, it makes upgrading easier.
2. I try to backpublish wherever I can (at least ScalaCheck and Discipline). Given the choice between "harder for me" and "harder for users", I pick "harder for me" :wink:

gabro: I'll try backpublishing. I've done it quite a few times in the past week for some internal projects, so I'm more confident than I usually would be in doing so :)

larsrh: If that works, I could backpublish all the way to Cats <3

julienrf:
Could you run the CI again?

larsrh:
It's not yet synced to Central.

gabro:
@julienrf I don't think it landed on Maven Central yet (it's on sonatype though!)

gabro:
The CI will still fail until Scalacheck is published, but Lars is on it

larsrh:
https://oss.sonatype.org/content/repositories/public/org/scalacheck/scalacheck_3/1.15.4/

gabro:
MUnit successfully backpublished 🎉 

https://oss.sonatype.org/content/repositories/public/org/scalameta/munit_3/

build.sbt
gabro: so, I've made a mistake backpublishing (feeling too confident, I knew it!) plus it turns out that backpublishing other typelevel libraries would be hard (due to old sbt-dotty versions), so I've restored RC3 support and I'll merge this and cut 0.7.26

tgodzik:
> ~MUnit successfully backpublished tada~
> 
> except I made a mistake and I've backpublished from master instead of using the 0.7.25 tag, so now MUnit 0.7.25 for Scala 3 uses Scalacheck 1.15.4, instead of 1.15.3. I'll cut a 0.7.26 release for both RC3 and stable

I think you might have also used JDK 11 by mistake, so it's not possible to use from 8.

gabro:
> I think you might have also used JDK 11 by mistake, so it's not possible to use from 8.

what's the issue with Java 8?

> Publishing by hand is evil 😅

I sinned, and I have been punished, I'll let the CI do its magic from now on

tgodzik (approved):
LGTM

tgodzik:
> > I think you might have also used JDK 11 by mistake, so it's not possible to use from 8.
> 
> what's the issue with Java 8?
> 

I tried to run mdoc tests for 3.0.0 and getting:
```
==> X tests.worksheets.WorksheetSuite.expected error running tests  0.003s java.lang.UnsupportedClassVersionError: munit/IgnoreSuite has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[error] Failed: Total 1, Failed 1, Errors 0, Passed 0
[error] Failed tests:
[error]         tests.worksheets.WorksheetSuite
[error] (worksheets / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 13 s, completed May 13, 2021 4:55:50 PM
```

gabro:
> I tried to run mdoc tests for 3.0.0 and getting:

I'll add a note to the release just in case 😭

PR #367, opened 2021-05-17
Create LICENSE

### LINKED ISSUE #365: Code license
I would like to re-use (part) of the code under `munit/shared/src/main/scala/munit/internal/difflib` -- I looked around for licensing information about it but could not find anything. Would it be ok for me to re-use that code freely? If you can, adding a license could be useful. :slightly_smiling_face:

gabro:
Closes #365 

Add an Apache 2.0 LICENSE file as already specified build.sbt at https://github.com/scalameta/munit/blob/main/build.sbt#L27-L29

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM!

olafurpg (commented):
LGTM thanks! Total oversight that this was missing

PR #376, opened 2021-06-07
Remove Node.js requirement with Scala.js

### LINKED ISSUE #247: Avoid relying on Node's fs when running in browser-based JSEnv
## Problem

MUnit relies on Node.js' `fs` module to read the source location when an assertion fails.

This is only possible if SJS if told to emit `CommonJSModule` or `ESModule`:

```scala
 scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule)),
```

**Issue** is that one cannot use such modules in browser-based JSEnv, filesystem doesn't exist there. And tests fail with an exception:

```scala
[error] munit.internal.JSFs$ needs to be imported from module 'fs' but module support is disabled
[error]   called from private java.nio.file.Files$.$anonfun$readAllBytes$1(java.nio.file.Path)[byte
[error]   called from java.nio.file.Files$.readAllBytes(java.nio.file.Path)[byte
[error]   called from java.nio.file.Files$.readAllLines(java.nio.file.Path)java.util.List
[error]   called from private munit.internal.console.Lines.$anonfun$formatLine$1(java.nio.file.Path)[java.lang.String
[error]   called from munit.internal.console.Lines.formatLine(munit.Location,java.lang.String,munit.Clues)java.lang.String
[error]   called from munit.internal.console.Lines.formatLine(munit.Location,java.lang.String)java.lang.String
```

## Proposed solution

@sjrd and @olafurpg propose a fallback solution, akin to

```scala
try { js.Dynamic.global.require("fs") } catch { ... }
```

to avoid having `fs` a hard dependency for MUnit.

### Reference material

* Gitter thread: https://gitter.im/scala-js/scala-js?at=5fa6f6c52a60f731f73fa8f1
* Same issue solved differently in Weaver Test: https://github.com/disneystreaming/weaver-test/pull/82/files

olafurpg:
Previously, it was not possible to use MUnit with browser-based
JavaScript environments because MUnit imported the Node.js `fs` module.
This commit refactors the code so that MUnit should gracefully recover
when the Node.js `fs` module is not available.

Fixes #247

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,8 +100,7 @@ val sharedJVMSettings: List[Def.Setting[_]] = List(
 ) ++ mimaEnable
 val sharedJSSettings: List[Def.Setting[_]] = List(
   skipIdeaSettings,
-  crossScalaVersions := allScalaVersions.filterNot(_.startsWith("0.")),
-  scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule))
+  crossScalaVersions := allScalaVersions.filterNot(_.startsWith("0."))
 )
 val sharedJSConfigure: Project => Project =
   _.disablePlugins(MimaPlugin)
--- a/munit/js/src/main/scala/java/io/File.scala
+++ b/munit/js/src/main/scala/java/io/File.scala
@@ -2,8 +2,8 @@ package java.io
 
 import java.net.URI
 import java.nio.file.Path
-
-import munit.internal.{JSIO, JSPath, NodeNIOPath}
+import munit.internal.JSIO
+import munit.internal.NodeNIOPath
 
 // obtained implementation by experimentation on the JDK.
 class File(path: String) {
@@ -55,8 +55,12 @@ class File(path: String) {
 object File {
   def listRoots(): Array[File] = Array(
     new File(
-      if (JSIO.isNode) JSPath.parse(JSPath.resolve()).root
-      else "/"
+      JSIO.path match {
+        case Some(p) => p.parse(p.resolve()).root.asInstanceOf[String]
+        case None    => "/"
+      }
+      // if (JSIO.isNode) JSPath.parse(JSPath.resolve()).root
+      // else "/"
     )
   )
 
@@ -64,10 +68,14 @@ object File {
     separator.charAt(0)
 
   def separator: String =
-    if (JSIO.isNode) JSPath.sep
-    else "/"
+    JSIO.path match {
+      case Some(p) => p.sep.asInstanceOf[String]
+      case None    => "/"
+    }
 
   def pathSeparator: String =
-    if (JSIO.isNode) JSPath.delimiter
-    else ":"
+    JSIO.path match {
+      case Some(p) => p.delimeter.asInstanceOf[String]
+      case None    => ":"
+    }
 }
--- a/munit/js/src/main/scala/java/nio/file/Files.scala
+++ b/munit/js/src/main/scala/java/nio/file/Files.scala
@@ -1,9 +1,10 @@
 package java.nio.file
 
+import scala.scalajs.js
 import java.{util => ju}
 import java.nio.charset.StandardCharsets
 
-import munit.internal.{JSFs, JSIO}
+import munit.internal.JSIO
 
 import scala.collection.JavaConverters._
 
@@ -13,8 +14,12 @@ object Files {
     val text = new String(bytes, StandardCharsets.UTF_8)
     text.linesIterator.toSeq.asJava
   }
-  def readAllBytes(path: Path): Array[Byte] = JSIO.inNode {
-    val jsArray = JSFs.readFileSync(path.toString)
+  def readAllBytes(path: Path): Array[Byte] = {
+    val jsArray = JSIO.fs match {
+      case Some(fs) =>
+        fs.readFileSync(path.toString).asInstanceOf[js.Array[Int]]
+      case None => new js.Array[Int](0)
+    }
     val len = jsArray.length
     val result = new Array[Byte](len)
     var curr = 0
--- a/munit/js/src/main/scala/munit/internal/JSIO.scala
+++ b/munit/js/src/main/scala/munit/internal/JSIO.scala
@@ -1,117 +1,38 @@
 package munit.internal
 
 import scala.scalajs.js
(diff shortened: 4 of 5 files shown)


munit/js/src/main/scala/java/io/File.scala
```diff
@@ -55,19 +55,27 @@ class File(path: String) {
...
   def listRoots(): Array[File] = Array(
     new File(
-      if (JSIO.isNode) JSPath.parse(JSPath.resolve()).root
-      else "/"
+      JSIO.path match {
+        case Some(p) => p.parse(p.resolve()).root.asInstanceOf[String]
+        case None    => "/"
+      }
+      // if (JSIO.isNode) JSPath.parse(JSPath.resolve()).root
+      // else "/"
```
gabro, line 63: leftover?

build.sbt
```diff
@@ -100,8 +100,7 @@ val sharedJVMSettings: List[Def.Setting[_]] = List(
 ) ++ mimaEnable
 val sharedJSSettings: List[Def.Setting[_]] = List(
   skipIdeaSettings,
-  crossScalaVersions := allScalaVersions.filterNot(_.startsWith("0.")),
-  scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule))
+  crossScalaVersions := allScalaVersions.filterNot(_.startsWith("0."))
```
tgodzik, line 103:
```suggestion
  crossScalaVersions := allScalaVersions
```
just as a complete side note, I don't think we need this anymore

raquo:
Thanks, checked it out locally. It does not throw as before, but now MUnit does not print any information about the assertion failures except the name of the test in which they appeared.

Munit on a JVM project prints this kind of useful error on failure:

```
==> X com.raquo.mapk.v1.V1MapKSpec.xxx  0.016s munit.ComparisonFailException: <projectRoot>/src/test/scala/com/raquo/mapk/v1/V1MapKSpec.scala:32
31:  test("xxx") {
32:    assertEquals(true, false)
33:  }
values are not the same
=> Obtained
true
=> Diff (- obtained, + expected)
-true
+false
```

Whereas MUnit in this PR prints only the following on a similar test in a JS project:

```
==> X com.raquo.domtestutils.MUnitJsDomSpec.xxx 0.01s munit.ComparisonFailException:
```

Also, the console is now printing a warning (not an error) before the tests:

```
[info] Writing and bundling the test loader
[warn] ./domtestutils-test-fastopt.js 4434:33-48
[warn] Critical dependency: the request of a dependency is an expression
[warn]     at CommonJsRequireContextDependency.getWarnings (<projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/webpack/lib/dependencies/ContextDependency.js:40:18)
[warn]     at Compilation.reportDependencyErrorsAndWarnings (<projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/webpack/lib/Compilation.js:1360:24)
[warn]     at <projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/webpack/lib/Compilation.js:1168:10
[warn]     at AsyncSeriesHook.eval [as callAsync] (eval at create (<projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:13:1)
[warn]     at AsyncSeriesHook.lazyCompileHook (<projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/tapable/lib/Hook.js:154:20)
[warn]     at Compilation.finish (<projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/webpack/lib/Compilation.js:1163:28)
[warn]     at <projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/webpack/lib/Compiler.js:622:17
[warn]     at eval (eval at create (<projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:9:1)
[warn]     at <projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/webpack/lib/Compilation.js:1095:12
[warn]     at <projectRoot>/target/scala-2.13/scalajs-bundler/test/node_modules/webpack/lib/Compilation.js:1007:9
[warn]     at processTicksAndRejections (internal/process/task_queues.js:79:11)
[warn]  @ ./domtestutils-test-fastopt-loader.js
```

You mentioned in previous comments that you need the `fs` module to read the source files to display information about the failure. So... does this PR replace that logic with a stub if the fs module can't be loaded?

build.sbt
gabro: 👍 I had a draft comment saying the same thing

olafurpg:
@raquo thank you for trying it out! What Scala.js linker settings are you using to reproduce that? It might require a small tweak to get the exception message

raquo:
@olafurpg Here is all of the linker config and a few more relevant bits:

```scala
scalaVersion := "2.13.5" // Scala.js version is 1.5.0
(Test / scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) })
(Test / requireJsDomEnv) := true
(installJsdom / version) := "16.4.0"
useYarn := true
(Test / parallelExecution) := false
```

No other `scalaJSLinkerConfig` settings (I also had source maps off, but commented that out, and still the same). I only added the line with `CommonJSModule` because MUnit docs told me to, but commenting it out does not change anything.

One weirdness on my side that I can think of is that my project where I'm trying to use this is a testing library, so it requires the MUnit dependency without appending `% "test"`. Whereas my `scalaJSLinkerConfig` has a `Test /` prefix as you see above. So maybe something is not catching on? I tried removing that prefix or replacing it with `Compile /`, but that did not produce any effect.

For the record, the branch that I'm trying to build is [here](https://github.com/raquo/scala-dom-testutils/tree/munit). The simple test that I made to fail is `MUnitJsDomSpec`. MUnit version is defined in Versions.sbt.

armanbilge:
@olafurpg I can confirm this PR works over at https://github.com/http4s/http4s/pull/4938.

Thanks for making this change, I'm badly in need of this. Is there anything I can do to get this merged and (snapshot) released ASAP? I'm happy to polish up this PR as @gabro and @tgodzik see fit.

PR #395, opened 2021-07-25
Improve error message when comparing arrays with same elements

### LINKED ISSUE #339: Better handling of Array[T] in assertions
Currently (version 0.7.23) munit doesn't provide special handling for comparing arrays, which has certainly tripped up users many times already.

uTest provides a [simple hack](https://github.com/com-lihaoyi/utest/blob/39e8d4cbc0b34f047c0da43c1013bc276bb73082/utest/src/utest/asserts/Asserts.scala#L82), which isn't pretty, but gets the job done.

The suggestion is to provide something similar (or even better) for special-casing arrays.

olafurpg:
Fixes #339.

### DIFF
--- a/munit/shared/src/main/scala/munit/Assertions.scala
+++ b/munit/shared/src/main/scala/munit/Assertions.scala
@@ -116,6 +116,28 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
   )(implicit loc: Location, ev: B <:< A): Unit = {
     StackTraces.dropInside {
       if (obtained != expected) {
+
+        (obtained, expected) match {
+          case (a: Array[_], b: Array[_]) if a.sameElements(b) =>
+            // Special-case error message when comparing arrays. See
+            // https://github.com/scalameta/munit/pull/393 and
+            // https://github.com/scalameta/munit/issues/339 for a related
+            // discussion on how MUnit should handle array comparisons.  Other
+            // testing frameworks have special cases for arrays so the
+            // comparison succeeds as long as `sameElements()` returns true.
+            // MUnit chooses instead to fail the test with a custom error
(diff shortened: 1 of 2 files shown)


munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -116,6 +116,28 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
   )(implicit loc: Location, ev: B <:< A): Unit = {
     StackTraces.dropInside {
       if (obtained != expected) {
+
+        (obtained, expected) match {
+          case (a: Array[_], b: Array[_]) if a.sameElements(b) =>
```
danicheg, line 121: That seems a bit puzzling to fail an assertion and ask the user to call the same assertion with converting `Array` -> `Seq`, although we have already done the job of checking elements... But I agree, it may help in understanding what's going on for those who forgot about `Array`'s equals checking. 

Shouldn't we add an analogical warning to `assertNotEquals` for a successful case (arrays are not equal by refs, but same on elements)?

danicheg (approved):
Glad to see improving the user experience for `Array`s checking.

munit/shared/src/main/scala/munit/Assertions.scala
olafurpg: I don’t think it’s puzzling. It’s just an error message in the style of “did you mean?” compile errors.

You can already achieve the assertNotEquals behavior with an assert. The main point of the custom assertions is to provide richer error messages for common usecases. You can also enrich the assert failure message with clues.

gabro (approved):
LGTM! After reading all the discussions, I think this is the approach more in line with the spirit of MUnit: good DX without the magic.

PR #397, opened 2021-08-06
Add buffered logging option

jenshalm:
This is the original commit which adds the new `+b` flag, plus a small bugfix and documentation.

IMHO the section "Run tests in parallel" should be on the Getting Started page, which focuses on running tests, and not on the page which focuses on writing tests. If you agree, I'm happy to make that change, too.

This PR does not contain any possible refactoring/simplification we discussed, as that would be better done with a separate PR.


olafurpg (approved):
LGTM 👍 

Feel free to follow up with a separate PR to enable this option by default. I think it makes sense to turn this on by default given that sbt uses parallel execution by default.

Issue #399, opened 2021-08-09
Concurrency issues in junit-interface module


jenshalm:
I believe some of the issues are also a concern in the original project the code was forked from,
however, I'll only describe them from the munit perspective.

**TLDR - my recommendations:**

1) Remove support for `--no-stderr` and `-q` (quiet) flags.
   They are very broken with parallel execution, which is the default in sbt, and cannot be made to work properly either.

2) Simplify `EventDispatcher` and `RichLogger` by removing their support for multi-suite and multi-thread usage,
   which is not needed by the munit implementation and which would have race conditions if those features were needed.

**In more detail:**

Regarding 1)

The `--no-stderr` and `-q` flags are currently not documented and also only supported for the JVM version of munit, 
but a user might stumble across them in the code and try them out which would lead to a pretty bad surprise.

These options redirect `System.out` and `System.err` respectively, and they do this on a per-suite basis.
This leads to unexpected behaviour in the default, parallel run mode, as the suites start to step on each other's toes.
In most cases this will leave `System.out` still pointing to a `ByteArrayOutputStream` even after all tests completed, permanently silencing munit until sbt is restarted. 
Even reverting the munit configuration change and using `reload` in the sbt console would not fix the issue. 
sbt has to be restarted to get back to normal, making this a really bad trap for users.

The issue is easy to reproduce. 

1) Set the `-q` flag with `Test / testOptions += Tests.Argument(TestFrameworks.MUnit, "-q")`.
2) Leave `parallelExecution` at its default (`true`)
3) Run `test` or `testOnly` with at least a handful of suites.

Outcome:

* 1st run: The vast majority of munit logs are swallowed, only some make it through.
* Every subsequent run: All munit logs are swallowed. Only sbt's own logger still works (they probably keep a reference to `System.out` internally).

Given that those features are 1) undocumented, 2) not supported for JS or Native and 3) essentially unfixable,
I think it is best to remove them.

Regarding 2) 

This is not a user issue, but more a maintainer's issue, so somewhat less critical.
I think the implementations of `EventDispatcher` and `RichLogger` are confusing as they suggest a use case
that does not exist.
In practice both are pinned to a single suite and a single thread (see `JUnitTask.execute`).
munit does not support nested suites or running the tests of a single suite in parallel.
The concurrent maps, keyed by suite, unnecessarily complicate the code, and give a false sense of security, 
as their implementation for concurrent use is not even correct.
[...]

PR #401, opened 2021-08-10
Concurrency fixes for junit-interface

jenshalm:
This PR attempts to address both issues in #399 (in separate commits).

Regarding 2): I only removed multi-suite support from `EventDispatcher` and `RichLogger`, not thread-safety.
Some basic observation suggests that both classes are pinned to a thread, even with asynchronous tests, but I wasn't 100% certain that this is the case in all possible scenarios. Most of the possible simplification is already achieved by dropping multi-suite support, so maybe keeping thread-safety does not hurt.


jenshalm:
The tests did not execute when I originally created the PR, I had to push a random change to actually trigger them. GitHub had issues earlier today, and apparently when checks don't execute during PR creation, they never will.

olafurpg (approved):
Thank you for this contribution! I apologize the slow review, I've been away on vacation. LGTM 👍

Issue #408, opened 2021-08-30
Loss of output when buffering is enabled


rossabaker:
I am inconsistently seeing successes, failures, and stack traces on CI under v0.7.28.  Running a specific test locally, I can reliably see that the new output buffering is hiding success messages on a simple test.

buffering on: https://github.com/http4s/http4s/commit/35c0d48c2029d3477f7444e28ad7e2097ba2163c:

```
org.http4s.MethodSuite:
[info] Passed: Total 32, Failed 0, Errors 0, Passed 32
[success] Total time: 8 s, completed Aug 30, 2021 5:20:39 PM
```

buffering off: https://github.com/http4s/http4s/commit/229e01e3c8c4e45a57d89811ddc358e28751aed2:

```
org.http4s.MethodSuite:
  + Hash[Method]: hash.antisymmetry eq 0.051s
  + Hash[Method]: hash.hash compatibility 0.009s
  + Hash[Method]: hash.reflexivity eq 0.004s
  + Hash[Method]: hash.same as scala hashing 0.004s
  + Hash[Method]: hash.same as universal hash 0.004s
...
[info] Passed: Total 32, Failed 0, Errors 0, Passed 32
[success] Total time: 4 s, completed Aug 30, 2021 5:19:05 PM
```

This is an enormous build.  I can try to minimize it if the cause is not immediately obvious.  I'm excited about the new feature, and loving the library in general!

olafurpg:
Thank you for reporting! I am able to reproduce the issue with a local clone of http4s at the commit https://github.com/http4s/http4s/commit/35c0d48c2029d3477f7444e28ad7e2097ba2163c I wasn't able to reproduce this inside the MUnit build but I took a blind stab at fixing this issue in #409. I'm unable to reproduce the issue in the http4s build after the fix so we can merge and release that commit

olafurpg:
I just cut a new 0.7.29 release that may take a few hours to become available on Maven Central

olafurpg:
0.7.29 is available now on Maven Central

rossabaker:
Thank you!  It's working well for me.

olafurpg:
Great! Thank you for confirming

PR #409, opened 2021-08-31
Attempt 1: make sure to always flush buffered logs

olafurpg:
Previously, buffered logs would sometimes not get flushed resulting in
lost test report output. See #408. This commit adds an extra attempt to
flush the buffered logs inside the sbt testing framework in case the
logs don't get flushed by the JUnit test listener. I'm not 100% sure
this fixes the problem but it seems worth a short and it shouldn't cause
a regression.


olafurpg:
cc/ @jenshalm

PR #413, opened 2021-09-10
Use provided `scalaJSVersion` and `nativeVersion`

lolgab:
Scala.js and Scala Native Sbt plugins already provide the versions for
the Sbt build to consume.


olafurpg (approved):
Thank you!

PR #418, opened 2021-09-19
Add asynchronous fixture

### LINKED ISSUE #186: Support async fixtures
Currently, `Fixture.{beforeEach/afterEach}` methods return `Unit`, which makes it impossible to implement those methods asynchronously. We should change the signature of those methods to also support returning a future.

danicheg:
Fixes #186.

This introduces `AsyncFixture[T]` that allows creating suite-local and test-local fixtures with asynchronous computations of `beforeXXX` and `afterXXX` methods. Also brings some refactoring to `MUnitRunner` internals to make it type-safer and easier to reason about, though more verbose.
I think it's better to have both `Fixture[T]` and `AsyncFixture[T]`. It's good to preserve compatibility when we can. Also for some cases, no need for asynchronous boundaries and `Fixture[T]` fits well as it is. But I'm not _so_ convinced about it and ready for discussion and any suggestions.


tests/jvm/src/test/scala/munit/TestLocalAsyncFixtureJVMSuite.scala
```diff
@@ -0,0 +1,33 @@
+package munit
+
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class TestLocalAsyncFixtureJVMSuite extends FunSuite {
```
danicheg, line 6: This suite was added for `JVM` and `JS` only because of the bug in native, see https://github.com/scala-native/scala-native/issues/1918

danicheg:
Have no idea why `scalafixCheckAll` was failed on CI - it_works_on_my_machine.jpg 😆 
Also, I've fixed the formatting. So, this is ready for review.

docs/fixtures.md
```diff
@@ -88,6 +88,38 @@ class FilesSuite extends FunSuite {
 }
 ```
 
+If `beforeEach()` and `afterEach()` methods supposed to be asynchronous computations
```
olafurpg, line 91: I prefer to use active sentences like "Use X" or "Override Y" when writing docs instead of "It's possible to use X" or "You can do Y".

```suggestion
Use `AsyncFixture[T]` for fixtures that need to be loaded asynchronously. Override the `munitAsyncFixtures` method to register async fixtures.
```

docs/fixtures.md
```diff
@@ -120,6 +152,40 @@ class MySuite extends munit.FunSuite {
 }
 ```
 
+If `beforeAll()` and `afterAll()` methods supposed to be asynchronous computations
```
olafurpg, line 155: Same comment as above about using active sentences like "Use X" or "Override Y".

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
-      if (isContinue) {
-        runAsyncTestsSynchronously(notifier)
-      } else {
-        Future.successful(())
+      Future.successful(())
+    } else {
+      val result =
+        for {
+          isBeforeAllRun <- runBeforeAll(notifier)
+          if isBeforeAllRun
```
olafurpg, line 152: What is the error message when this condition fails? It would be best to fail with an exception that explains the reason why tests didn't run: "Aborting test execution because `beforeAll` hooks failed to run" (or something like that).

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
 
-  private def runBeforeAll(notifier: RunNotifier): Boolean = {
-    var isContinue = runHiddenTest(notifier, "beforeAll", suite.beforeAll())
-    suite.munitFixtures.foreach { fixture =>
-      isContinue &= runHiddenTest(
-        notifier,
-        s"beforeAllFixture(${fixture.fixtureName})",
-        fixture.beforeAll()
+  private def runBeforeAll(notifier: RunNotifier): Future[Boolean] = {
+    val suiteBeforeAllF = Future(
```
olafurpg, line 170: We should ensure that all-synchronous test suites continue to run synchronously. Stack traces become difficult to reason about in async code

```suggestion
    val suiteBeforeAllF = Future.successful(...
    // OR
    val suiteBeforeAllF = Future.fromTry(...
```

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
+      runHiddenTest(notifier, "beforeAll", suite.beforeAll())
+    )
+
+    def beforeAllFixtureF =
+      Future.sequence(
+        suite.munitFixtures.map(f =>
+          Future(
```
olafurpg, line 177: Same comment as above about using `Future.successful` or `Future.fromTry`

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
-        notifier,
-        s"beforeAllFixture(${fixture.fixtureName})",
-        fixture.beforeAll()
+  private def runBeforeAll(notifier: RunNotifier): Future[Boolean] = {
+    val suiteBeforeAllF = Future(
+      runHiddenTest(notifier, "beforeAll", suite.beforeAll())
+    )
+
+    def beforeAllFixtureF =
+      Future.sequence(
```
olafurpg, line 175: We should use a tail-recursive method to evaluate non-async futures in sequence so that stack traces doesn't get huge. If you're using a blocking `ExecutionContext` the stack trace will grow linearly with the number of test cases in the test suite. It should be possible to write test suites with ~100k test cases.

See `runAsyncTestsSynchronously` for how to do this. The trick is to have special handling when `future.value` is non-empty. You can't use the `@tailrec` annotation because the async path cannot be optimized for tail recursion, but the compiler still optimizes the sync path even if the annotation is not present.

https://github.com/scalameta/munit/blob/6484bf798c5fd3851e9b919964ab4d0837c57e94/munit/shared/src/main/scala/munit/MUnitRunner.scala#L127-L140

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
+            )
+          )
+        )
       )
-    }
-    isContinue
+
+    def beforeAllAsyncFixtureF =
+      Future
+        .sequence(
```
olafurpg, line 189: Same comment as before, we should avoid using `Future.sequence`

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
-        s"afterAllFixture(${fixture.fixtureName})",
-        fixture.afterAll()
+
+  private def runAfterAll(notifier: RunNotifier): Future[Unit] = {
+    def suiteAfterAllF = Future(
+      runHiddenTest(notifier, "afterAll", suite.afterAll())
+    )
+
+    def afterAllFixtureF =
+      Future.sequence(
```
olafurpg, line 215: Same as above.

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
   }
-  private def runAfterAll(notifier: RunNotifier): Unit = {
-    suite.munitFixtures.foreach { fixture =>
-      runHiddenTest(
-        notifier,
-        s"afterAllFixture(${fixture.fixtureName})",
-        fixture.afterAll()
+
+  private def runAfterAll(notifier: RunNotifier): Future[Unit] = {
+    def suiteAfterAllF = Future(
```
olafurpg, line 210: Same as above, no `Future.apply`

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
+      runHiddenTest(notifier, "afterAll", suite.afterAll())
+    )
+
+    def afterAllFixtureF =
+      Future.sequence(
+        suite.munitFixtures.map(f =>
+          Future(
```
olafurpg, line 217: No `Future.apply`

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
+              f.afterAll()
+            )
+          )
+        )
       )
-    }
-    runHiddenTest(notifier, "afterAll", suite.afterAll())
+
+    def afterAllAsyncFixtureF =
+      Future.sequence(
```
olafurpg, line 228: Same as above.

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -144,80 +144,215 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
+
+    for {
+      _ <- afterAllFixtureF
+      _ <- afterAllAsyncFixtureF
+      _ <- suiteAfterAllF
+    } yield ()
   }
 
-  class BeforeEachResult(
+  private[munit] class BeforeEachResult(
```
olafurpg, line 247: 👍

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -279,61 +414,85 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
...
           }
         }
-        scala.util.Failure(head)
-      case _ =>
-        scala.util.Success(())
-    }
-  }
+      )
+
+      aggregated = computed.foldLeft(List.empty[Throwable] -> List.empty[A]) {
```
olafurpg, line 476: We should avoid methods like `foldLeft` since they increase the stack trace. It's better to write out the recursion by hand with `@tailrec`

olafurpg (changes requested):
Thank you for this contribution! I have wanted to support async fixtures for a long time so I'm fully on board with adding this new feature. I like that you added a separate API to avoid breaking changes.

The refactoring in `MUnitRunner` is fairly involved so we may need several iterations to get the code into a state that's ready for merging. I did a quick review through the GitHub web UI and I'd love to do at least one review in my IDE and run tests locally to make sure that everything is working as expected. I've tried to make the test suite as comprehensive as possible but some behavior is currently not covered by tests (like stack safety on sync methods).

Don't hesitate to oppose my comments, I may misunderstand something how `Future.sequence` works under the hood. My comments are based on my understanding of how things work, but that understanding may be wrong or outdated!

munit/shared/src/main/scala/munit/MUnitRunner.scala
olafurpg: My bad, `List.foldLeft` doesn't have this problem. I still prefer manually written tailrec methods over `foldLeft`

munit/shared/src/main/scala/munit/MUnitRunner.scala
danicheg: Shame on me. Nice catch, though! I've rewritten it with preserving the semantic of the original version.

munit/shared/src/main/scala/munit/MUnitRunner.scala
danicheg: Agree.

munit/shared/src/main/scala/munit/MUnitRunner.scala
danicheg: Seems reasonable to me. I've added a similar to `runAsyncTestsSynchronously` recursion method for execution of a bunch of Futures.

munit/shared/src/main/scala/munit/MUnitRunner.scala
danicheg: Done.

munit/shared/src/main/scala/munit/MUnitRunner.scala
danicheg: Done.

munit/shared/src/main/scala/munit/MUnitRunner.scala
danicheg: I'm in love with `fold`s. But for keeping consistency in the codebase have rewritten it to the tail-recursion function.

danicheg:
@olafurpg thank you for that detailed review. It'd be great to take another iteration on this.

olafurpg:
@danicheg I apologize the long delay. I spent a while trying to review and refactor the code in this PR but I wasn't happy with the overall changes. I ended up trying a different approach that I like better https://github.com/scalameta/munit/pull/430. This different approach allows us to support async and sync fixtures with the same API while removing several cryptic types from the public API (`GenericTest[T]` and friends) and allowing toplevel classes to implement fixtures. This approach is a breaking change but I think it's worth it considering the benefits.

olafurpg:
Superseded by #430, which adds support for async fixtures using a different approach. Thank you @danicheg for this contribution! Even if we didn't end up merging this PR it was very influential in getting async fixtures implemented

PR #427, opened 2021-10-05
More correct string inequality error message

### LINKED ISSUE #152: Fail exception report is incorrect
I get the following exception with the code that follows:
`values are not equal even if they have the same `toString()`: bb as,dn`

Longer version:
```
==> X org.scalajs.testsuite.javalib.util.PropertiesSuite.string leading  0.001s munit.FailException: /Users/eric/workspace/portable-scala-java-test/testSuite/shared/src/test/scala/org/scalajs/testsuite/javalib/util/PropertiesSuite.scala:162 values are not equal even if they have the same `toString()`: bb as,dn
161:    println(s"'$b'=='$res1' ${b==res1}")
162:    assertEquals(b, res1)
163:  }
```
Here are the tests:
```Scala
val b = "bb as,dn"
  test("string trailing"){
    val res0 = "bb as,dn   "
    println(s"'$res0'=='$b' ${res0==b}")
    assertEquals(res0, b)
  }

  test("string leading") {
    val res1 = "   bb as,dn"
    println(s"'$b'=='$res1' ${b==res1}")
    assertEquals(b, res1)
  }
```

raboof:
The current logic may claim values have the same `toString()`
when they don't. The 'full' solution would of course be to
have a 'strict' string diff, but until then let's at least
not lie about it.

Fixes #152

### DIFF
--- a/munit/shared/src/main/scala/munit/Assertions.scala
+++ b/munit/shared/src/main/scala/munit/Assertions.scala
@@ -153,11 +153,18 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
           munitPrint(clue),
           printObtainedAsStripMargin = false
         )
-        failComparison(
-          s"values are not equal even if they have the same `toString()`: $obtained",
-          obtained,
-          expected
-        )
+        if (obtained.toString() == expected.toString())
+          failComparison(
+            s"values are not equal even if they have the same `toString()`: $obtained",
+            obtained,
+            expected
+          )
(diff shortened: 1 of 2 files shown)


olafurpg (approved):
LGTM 👍 Thank you! I agree the new message is better

som-snytt:
Just getting (re-)acquainted with munit today, and I needed this fix so bad. My expected was missing a trailing newline.

PR #430, opened 2021-10-10
Add support for async fixtures

olafurpg:
Previously, it was not possible to create fixtures that loaded
asynchronously. It was possible to work around this limitation on the
JVM by awaiting on futures, but there was no workaround for Scala.js.
This commit adds support to return futures (and anything that converts
to futures) from the beforeAll/beforeEach/afterEach/afterAll methods.

Supersedes https://github.com/scalameta/munit/pull/418. This commit is
inspired by that PR but uses a different approach:

- No new `AsyncFixture[T]` type. This simplies the logic in
  `MUnitRunner` and reduces the size of the MUnit public API.
  The downside is that we introduce a breaking change to the existing
  `Fixture[T]` API, which will require bumping the version to v1.0.0
  for the next release.
- This approach supports any `Future`-like values by hooking into the
  default evaluation of test bodies via `munitValueTransforms`.

Co-authored-by: Daniel Esik <e.danicheg@yandex.ru>


olafurpg:
I'll try to get back to this PR soon so that it doesn't stay open for too long.

olafurpg:
@gabro I considered it but didn't want to write `List[Fixture[Id, String]]` or `List[Fixture[Future, String]]` for common cases. Under the hood, MUnit works with `Any` and `Future[Any]` so it doesn't buy us cleaner internal code. For end users, I feel like `Fixture[String]` and `FutureFixture[String]` and `TaskFixture[String]` will be more ergonomic to write and it will result in more readable code.

olafurpg:
For the record, MUnit has abstracted on the container type `Future[_]` from the start and it's been possible to write custom `Suite` classes for other containers like `IO`. The idea was that `munit.FunSuite` was special cased for `Future[_]` but you could implement test suites based on another container. This PR removes that functionality because MUnit internals have always been tied to `Future` anyways and most users seemed fine with using `munit.FunSuite` directly (even to integrate with cats-effect IO).

olafurpg:
It's a bit confusing, but `IOFixture` would need to extend `AnyFixture` since the lifecycle method signatures should return `IO[Unit]`. Value transforms would take care of converting those IO values into futures.

armanbilge:
> Fun fixtures are test-local only and require you to re-define the fixture variable name for every test,

Yes, I'm aware, sorry for the confusion :) I linked that example because it shows the conversion to `Future`. There's also a fixture implementation that is JVM-only since it replies on a blocking call:

https://github.com/typelevel/munit-cats-effect/blob/main/common/jvm/src/main/scala/munit/CatsEffectFixturesPlatform.scala

olafurpg:
It would be a good exercise to implement the new Fixture API with `Resource`. I know some people have achieved it on the JVM with blocking I/O and it's not pretty. The code won't be much prettier with this PR since `Fixture[_]` is inherently a state-ful API, but it will at least avoid the blocking I/O making it possible to use MUni async fixtures on Scala.js.

armanbilge:
Thanks! Really excited about this.

Issue #431, opened 2021-10-14
support ESModule?


bblfish:
The docs say to use scala-js one needs to add this line to `build.sbt`
```scala
Test / scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
```
Would it be problematic to have this work?
```scala
Test / scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.ESModule)}
```

I was thinking producing ESModules would be better, though I am not that sure, as I am new to scala-js.

armanbilge:
Yes, this is not necessary anymore 👍

bblfish:
yes, [the doc](https://scalameta.org/munit/docs/getting-started.html) actually says so I guess, in a very minimal way.

```scala
// The use of "Test / " allows the rest of your project to use a different module kind
Test / scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
```

But my project is quite complex, so I guess I am encountering some other problems that are making things a bit more complicated... But

armanbilge:
Even that is technically outdated, that line is not necessary at all anymore for `Test` or otherwise.

olafurpg:
Any pull requests to clarify the Scala.js docs are welcome 🙏

PR #438, opened 2021-10-16
Remove Scala.js-specific instructions

armanbilge:
I can confirm this is no longer necessary as of #376.

Following up on #431.


olafurpg (approved):
Thank you!

PR #441, opened 2021-10-17
Move JDK shims to `munit.internal`

### LINKED ISSUE #437: Remove Scala.js shims for JDK classes
Specifically: https://github.com/scalameta/munit/tree/main/munit/js/src/main/scala/java

I know this will be an annoying change to make, but it would be great if this could land in 1.0 as well. I understand these shims are important for implementing munit features, but they are also potential landmines.

Here are two ways it can go wrong (and has for me):
1. I write a Scala.js library accidentally using `File` or `Path` or one of the other classes shimmed here. I write an munit suite to test it, and voila it all works (b/c munit has kindly put these JDK shims on the test classpath). So I publish my library. Then, a downstream depends on my library and tries to create a Scala.js app ... but they can't, because they get linking errors for `Path`, `File`, etc. Having JDK shims on the test classpath makes it nigh impossible for libraries to catch these sorts of mistakes.
2. If some other well-meaning library or application has _also_ shimmed these JDK classes, then there would be a classpath conflict, in which case I'm actually not sure what happens. But probably not good.

Off the top of my head, there's a couple ways to avoid this (not mutually exclusive):
1. Move these shims out of the `java` package into the `munit` package and in your JVM sources create type aliases to point e.g. `munit.internal.Path` to the JDK implementations.
2. Create wrapper classes in the `munit` package, that delegate to JDK on the JVM and Node.js on JS.

Thanks for your attention to this issue.

armanbilge:
Will fix #437.


munit/js/src/main/scala/munit/internal/Path.scala
```diff
@@ -1,70 +1,63 @@
...
-import java.nio.file.WatchEvent.{Kind, Modifier}
-import java.nio.file._
 import java.util
 
 import scala.collection.JavaConverters._
 
 // Rough implementation of java.nio.Path, should work similarly for the happy
 // path but has undefined behavior for error handling.
-case class NodeNIOPath(filename: String) extends Path {
+case class Path(filename: String) {
```
olafurpg, line 10: Let's call this `MUnitPath` instead of `Path`. Or something else if you have suggestions. I don't like it when libraries use generic names for internal classes because it pollutes the namespace when you're auto-completing symbols.

munit/js/src/main/scala/munit/internal/Paths.scala
```diff
@@ -1,10 +1,7 @@
-package java.nio.file
+package munit.internal
```
olafurpg, line 1: Let's introduce a new `munit.internal.io` package. The general structure I try to following is

- flat list of classes/objects inside `munit._` for public API
- flat list of packages inside `munit.internal._`

The reason is mostly historical because MUnit includes a reimplementation of the java difflib and it didn't feel right to have all of those diff classes inside `munit.internal._`.

munit/jvm/src/main/scala/munit/internal/package.scala
```diff
@@ -0,0 +1,13 @@
+package munit
+
+package object internal {
+  type File = java.io.File
```
olafurpg, line 4: Can we add an object called `PlatformIO` that has these methods and aliases instead? I try to avoid package objects when possible.

olafurpg (changes requested):
Thank you for this contribution! Looks good to me overall, just a few nitpick comments on undocumented conventions I try to keep in the codebase.

olafurpg:
You can run `scalafixAll` and `scalafmtAll` inside the sbt shell to fix the Scalafix error

armanbilge:
Thanks for the feedback and pointers. I'm still tracking down various linking issues for JS then I'll iterate on your feedback 👍

armanbilge:
Uh-oh, maybe I should wait on this until you decide what you want to do? 😅  Removing I/O completely would be excellent for browsers and friends.

olafurpg:
You can try to implement this alternative approach if you're interested. The steps would be

1. Add new `textContents` field to `munit.Location`, with a fallback constructor that defaults to `textContents = ""`  https://github.com/scalameta/munit/blob/e331ee3666838edb6318653c7d903a6e193e4be2/munit/shared/src/main/scala/munit/Location.scala#L14 
2. Update the Scala 2 macro to inline the contents here https://github.com/scalameta/munit/blob/e331ee3666838edb6318653c7d903a6e193e4be2/munit/shared/src/main/scala/munit/internal/MacroCompatScala2.scala#L15
3. Update the Scala 3 macro to inline the contents here https://github.com/scalameta/munit/blob/e331ee3666838edb6318653c7d903a6e193e4be2/munit/shared/src/main/scala-3/munit/internal/MacroCompat.scala#L20
4. Update `Lines.scala` to use `Location.textContents` instead of file I/O

olafurpg:
I'm sorry for guiding you in the other direction, this idea didn't come up until after you opened the draft PR.

olafurpg:
The CI seems green so we can also merge this PR as is and I can look into the alternative later.

armanbilge:
Apologies, getting a bit out-of-scope for me. As it stands this PR does the hard part of #437 by tracking down all the JDK stuff and condenses it. All that's left is bike-shedding names according to unwritten convention which it seems you are the expert on 😉

armanbilge:
Apologies, I regret my remark as it came across poorly I think 😕 I have ton of respect for projects and developers that keep things organized, extremely important for projects like this at the top of the dependency food chain. It's just hard for me to know exactly how you want things without a couple more iterations and it seems the future of this PR is dubious depending on some very good changes that might be forthcoming. So I'm happy to let this sit and if it's still relevant by the time 1.0 is solidifying then we can make it all perfect :)

armanbilge:
@olafurpg we've been butting heads with this problem again in https://github.com/http4s/http4s/pull/5758.

Would you be open to accepting the changes in this PR into 0.7.x series (after it's polished up)? I know it's technically a breaking change, but in fact anyone who experiences that breaking change likely had bugs/linking issues that were masked precisely because of these shims.

Thanks!

armanbilge:
Oh, one more idea: if you really want to avoid a breaking change, we apply the changes here, and also move these shims into a new artifact. So anyone that experiences issues because of the change, can just add a new dependency to get the shims back (or alternatively, the shims can be included by default and downstreams where it causes a problem can remove it explicitly).

munit/js/src/main/scala/munit/internal/Path.scala
armanbilge: Makes sense, done.

munit/js/src/main/scala/munit/internal/Paths.scala
armanbilge: Done.

munit/jvm/src/main/scala/munit/internal/package.scala
armanbilge: I moved the io-related ones to `PlatformIO`. I moved the reflect ones to `PlatformCompat`.

armanbilge:
@olafurpg I've now renamed/re-organized everything as you suggested. Thanks for your patience on this one! :)

olafurpg (approved):
LGTM thank you!

PR #443, opened 2021-10-18
Update website to point to latest stable release

olafurpg:


munit/shared/src/main/scala/munit/Test.scala
```diff
@@ -52,4 +52,5 @@ final class Test(
     buf += location
     buf.toArray
   }
+  l
```
gabro, line 55: leftover?

build.sbt
```diff
@@ -330,7 +330,8 @@ lazy val docs = project
       (ThisBuild / baseDirectory).value / "website" / "target" / "docs",
     mdocExtraArguments := List("--no-link-hygiene"),
     mdocVariables := Map(
-      "VERSION" -> version.value.replaceFirst("\\+.*", "")
+      "VERSION" -> version.value.replaceFirst("\\+.*", ""),
+      "STABLE_VERSION" -> "0.7.29"
```
gabro, line 334: I would've probably left `VERSION` as-is, and introduced a `NEXT_MAJOR_VERSION` or something, so that you wouldn't have to change existing uses of `VERSION`, but it's a minor thing, feel free to ignore

PR #461, opened 2021-11-26
Enable -Xlint and -Werror

som-snytt:
Linting on 2.13 such as required empty args,
inferred Any in sameElements of Array[_],
Timeout is double-valued, avoid Any.equals,
some elements unused, deprecations.
Allow JavaConverters.

Can't use nowarn and unused annotations
under 2.11/2.12.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ def scala213 = "2.13.13"
 
 def scala212 = "2.12.19"
 
-def scala3 = "3.1.2"
+def scala3 = "3.3.3"
 def junitVersion = "4.13.2"
 def gcp = "com.google.cloud" % "google-cloud-storage" % "2.32.1"
 inThisBuild(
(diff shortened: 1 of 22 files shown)


olafurpg (approved):
LGTM 👍 Thank you for this contribution! I pushed an empty commit to trigger the CI.

olafurpg (changes requested):
Is there a way to silence the new warnings when compiling the codebase? I don't like fatal warnings because they are painful to work with when coding locally (example: commenting out a println statement introduces a fatal unused warning). 

image

som-snytt:
Thanks for taking a look.

OK, less-fatal warnings. Ideally there would be a turn-key or push-button way to enable that only for continuous integration later in the workflow.

I see I added some `-Wconf`, maybe I forgot to clean-compile. This is only my second time using `-Wconf` outside testing `-Wconf`.

som-snytt:
Tweaked the scalafix task to use current 2.12. That didn't help warnings, so 2.12. scalac options avoid warning.

Now I've lost the doc command that failed, so I haven't reproduced that.

som-snytt:
Further exploration of perils of crossbuilding with lints.

PR #464, opened 2021-12-01
Draft: Change test return types from Any to overridable type MunitRes <: Any

lorandszakacs:
This is really a starting point for a discussion, but that has real code to showcase :D I'll hapilly move it to an issue.

The idea behind this is to let libraries that build on top of `munit` to constrain the return type of tests, particularly useful would be for uses in `munit-cats-effect` where one could then define a base class for cats-effect tests like:

```scala
import munit._
import cats.effect._

abstract class CatsEffectFunSuite extends BaseFunSuite {
  override type MunitRes = IO[Unit]
}
```

As far as I can tell this change would impact other libraries, and not end users.

### DIFF
--- a/munit/shared/src/main/scala/munit/FunFixtures.scala
+++ b/munit/shared/src/main/scala/munit/FunFixtures.scala
@@ -7,6 +7,7 @@ import scala.util.Success
 import scala.util.Failure
 
 trait FunFixtures { self: BaseFunSuite =>
+  override type MunitRes = Any
 
   class FunFixture[T] private (
       val setup: TestOptions => Future[T],
@@ -20,12 +21,12 @@ trait FunFixtures { self: BaseFunSuite =>
       )
 
     def test(name: String)(
-        body: T => Any
+        body: T => MunitRes
     )(implicit loc: Location): Unit = {
       fixture.test(TestOptions(name))(body)
     }
     def test(options: TestOptions)(
-        body: T => Any
+        body: T => MunitRes
     )(implicit loc: Location): Unit = {
       self.test(options) {
         implicit val ec = munitExecutionContext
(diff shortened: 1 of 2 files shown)


som-snytt:
Another use case mentioned in a thread: test code produces a `Try` which failed. User could provide a conversion that `_.get` to fail the test.

olafurpg:
Thank you for this contribution and I apologize for the delayed response. I'm on the fence about introducing this change because it makes the MUnit API less self-explanatory for the majority of users that won't customize the type alias. A regular user would encounter this name `MUnitRes` (or any other name we come up with) in their IDE every time they call `test()` and would need to remember that this type is in fact `Any`. The API will also be confusing for users who customize this type to be `IO[_]` because they will also need to remember that `MUnitRes` is `IO[_]`.

As a workaround, have you considered implementing a custom `ValueTransform` to fail the test if a non-IO value is returned?

```scala
    new ValueTransform(
      "IO",
      {
        case e: IO[_] => e.toFuture()
        case other        => sys.error(s"non-IO value $other") // report some helpful error message
      }
    )
```

This workaround won't catch mistakes at compile-time but that's maybe OK because you run the tests anyways. I'm fully aware that it's not idiomatic Scala to catch type errors at runtime but I feel like it may be an acceptable tradeoff in the special case of the `test() {}` body. If you are fully invested in the cats-effect ecosystem then you may prefer to use a more dedicated testing framework like weaver-test (https://disneystreaming.github.io/weaver-test/) that takes advantage of the awesome properties of `IO`, for example clean parallelization within a test suite.

olafurpg:
> Another use case mentioned in a thread: test code produces a `Try` which failed. User could provide a conversion that `_.get` to fail the test.

@som-snytt Have you tried using `ValueTransform`? It's specifically designed for cases like this.

olafurpg:
I just found this related PR to the MUnit integration for cats-effect, which fails the test at runtime for nested `IO[IO[_]]` values. The munit-cats-effect maintainers may be open to add an option to automatically include a value transform that also fails the tests for non-IO values (with maybe an exception for `Unit`)  https://github.com/typelevel/munit-cats-effect/pull/160

lorandszakacs:
I completely forgot about this :see_no_evil:

I understand the confusion introduced by some weird type member. It was more of a curiosity of how we can provide both tools for compile-time customization for downstream users in addition to runtime ones.

PR #477, opened 2022-01-05
Cross publish native

kpodsiad:
An attempt to cross publish munit-native & rest to Scala 3. 

Needs Scala 3.1.1 to be released, otherwise [macro mixing](https://docs.scala-lang.org/scala3/guides/migration/tutorial-macro-mixing.html) in [MacroCompat2](https://github.com/scalameta/munit/blob/dd5d392143c463d58c0d9ca06917169ec772d959/munit/shared/src/main/scala/munit/internal/MacroCompatScala2.scala) won't work. It fails with the `Experimental erased may only be used with a nightly or snapshot version of the compiler` - see https://github.com/lampepfl/dotty/pull/13795 for more details.


build.sbt
```diff
@@ -6,7 +6,7 @@ def previousVersion = "0.7.0"
 def scala213 = "2.13.6"
 def scala212 = "2.12.15"
 def scala211 = "2.11.12"
-def scala3 = "3.0.1"
+def scala3 = "3.1.0"
```
gabro, line 9: This will force munit users to upgrade their Scala 3 version as well :/ not sure there's a better way though

build.sbt
```diff
@@ -126,18 +126,21 @@ lazy val mimaEnable: List[Def.Setting[_]] = List(
...
   skipIdeaSettings,
   crossScalaVersions := allScalaVersions.filterNot(_.startsWith("0."))
 )
 val sharedJSConfigure: Project => Project =
   _.disablePlugins(MimaPlugin)
+
 val sharedNativeSettings: List[Def.Setting[_]] = List(
   skipIdeaSettings,
-  crossScalaVersions := scala2Versions
+  crossScalaVersions := allScalaVersions
```
gabro, line 143: I think crossScalaVersions can be now moved to sharedSetting since it's the same for all platforms (the filter in the JS version is a leftover of the Dotty 0.x days)

munit/non-jvm/src/main/scala/munit/internal/junitinterface/CustomFingerprint.scala
```diff
@@ -4,8 +4,9 @@ import sbt.testing.SubclassFingerprint
 
 class CustomFingerprint(
     val suite: String,
-    val isModule: Boolean
+    _isModule: Boolean
```
gabro, line 7: What's the deal with all these underscores changes?

build.sbt
kpodsiad: I set 3.1.0 because Scala Native 0.4.3-RC1 was released only for this version. However, `3.1.1` is still needed for macro mixing and 3.1.1 should have `scala-target` (I think that's how they called it) compiler flag which will allow targeting `3.0.2`.

munit/non-jvm/src/main/scala/munit/internal/junitinterface/CustomFingerprint.scala
kpodsiad: In scala3 you can't omit parens if the method was declared with them. It also applies to overriding by vals.
Check this example - https://gist.github.com/kpodsiad/8242ad1ff9eb959f3632dc63e3965f03. It will yield
```
[error]   Hello.scala:8:16: error overriding method name in trait Framework of type (): String;
[error]   value name of type String has incompatible type
```

build.sbt
gabro: Aha, got it. Maybe we can set 3.1.1 for the native build only and build with 3.0 for the other targets? Otherwise it's a breaking change for jvm and js users

munit/non-jvm/src/main/scala/munit/internal/junitinterface/CustomFingerprint.scala
gabro: Ok, was this enforced in 3.1? Because apparently it was working fine in 3.0

gabro: Ah no, wait, these files were never compiled with scala 3, got it

build.sbt
kpodsiad: If I understand correctly there shouldn't be any breaking changes if we'll set `-scala-release 3.0` (that's how it'll be called)

gabro: Is that available already? I thought it was still in the works

gabro: Sorry, you mentioned it in the first comment and I totally missed it 😫 it's late for me I suppose

kpodsiad: `-scala-release` hasn't been release in `3.1.1` and I decided to use your advice, I set `3.1.1` only for native build.

project/plugins.sbt
```diff
@@ -3,10 +3,13 @@ addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.10.0")
...
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.3")
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.31")
+
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.1.0")
-addSbtPlugin("org.portable-scala" % "sbt-scala-native-crossproject" % "1.1.0")
-addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.4.0")
 addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.6.0")
+
+addSbtPlugin("org.portable-scala" % "sbt-scala-native-crossproject" % "1.1.0")
+addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.4.3-RC2")
```
sideeffffect, line 11: We can drop the `-RC2` now :smiley:

kpodsiad:
@sideeffffect @gabro I'll answer to https://github.com/scalameta/munit/pull/488 here. There's a problem with scalacheck - there is no scalacheck published for Scala 3 & Native 0.4.x, see [maven](https://mvnrepository.com/artifact/org.scalacheck/scalacheck). I tried various tricks with `CrossVersion.for3Use2_13` and etc. but without any results.
When running `++3.1.1 testsNative/test` I always ended with `munit` depending on `0.4_3` and `munitScalacheck` depending on the `0.4_2.13` which unfortunately doesn't work.

@sideeffffect if you want you can play with it to get `++3.1.1 testsNative/test` passing. I'm not a multiversion multiplatform expert, maybe you can get it done ;)

There is a PR to support scala-native Scala 3 in scalacheck repo https://github.com/typelevel/scalacheck/pull/868. Unfortunately, it's waiting for maintainers to look at it.

kpodsiad:
I see 3 options:
- find a workaround for `scalacheck` scala3 & Native 0.4.x and made `++3.1.1 testsNative/test` passing. Can give it another try but it won't be in this week :/
- wait for typelevel/scalacheck#868
- extract scalacheck related tests to separate module, then `munit` can be released without any hassle and `munit-scalacheck` will wait for typelevel/scalacheck#868

@gabro Are you in favor of any option?

gabro:
Well if we have a viable workaround, sure, but it really depends on whether you or someone else can find it.

Reaching out to Typelevel about the status of Scalacheck and getting that PR merged upstream would be the preferable option for sure, while I'm not really keen on the idea of releasing MUnit without full Scalacheck support.

sideeffffect:
> I tried various tricks with `CrossVersion.for3Use2_13` and etc. but without any results.
> When running `++3.1.1 testsNative/test` I always ended with `munit` depending on `0.4_3` and `munitScalacheck` depending on the `0.4_2.13` which unfortunately doesn't work.

I think I've also hit this issue
```scala
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/home/ondra/Projects/sbt-locales/"), "apiNative"):
[error]    org.scala-native:windowslib_native0.4 _3, _2.13
...
```
Please read more here:
https://github.com/portable-scala/portable-scala-reflect/issues/23#issuecomment-1021617780

avdv:
Note, their is also a problem with scala-native itself you would need to workaround: https://github.com/scala-native/scala-native/issues/2543

sideeffffect:
For a workaround to the 
```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/home/ondra/Projects/sbt-locales/"), "apiNative"):
[error]    org.scala-native:windowslib_native0.4 _3, _2.13
```
issue, see https://github.com/cquiroz/sbt-locales/pull/190

avdv:
Using an inline conditional to work around issue scala-native/scala-native#2543 would be possible:
```
diff --git a/munit/native/src/main/scala/munit/internal/PlatformCompat.scala b/munit/native/src/main/scala/munit/internal/PlatformCompat.scala
index c08636e..e625d88 100644
--- a/munit/native/src/main/scala/munit/internal/PlatformCompat.scala
+++ b/munit/native/src/main/scala/munit/internal/PlatformCompat.scala
@@ -38,7 +38,7 @@ object PlatformCompat {
 
   def isJVM: Boolean = false
   def isJS: Boolean = false
-  def isNative: Boolean = true
+  inline val isNative = true // Scala 3 syntax
 
   def newRunner(
       taskDef: TaskDef,
diff --git a/munit/shared/src/main/scala-3/munit/internal/Compat.scala b/munit/shared/src/main/scala-3/munit/internal/Compat.scala
index 7e489e2..92b507d 100644
--- a/munit/shared/src/main/scala-3/munit/internal/Compat.scala
+++ b/munit/shared/src/main/scala-3/munit/internal/Compat.scala
@@ -7,7 +7,10 @@ object Compat {
   val LazyList = scala.LazyList
   def productElementNames(p: Product): Iterator[String] =
     p.productElementNames
-  def collectionClassName(i: Iterable[_]): String = {
-    i.asInstanceOf[{ def collectionClassName: String }].collectionClassName
+  inline def collectionClassName(i: Iterable[_]): String = {
+    if PlatformCompat.isNative then
+        i.getClass.getName // or just "iterable" ?
+    else
+        i.asInstanceOf[{ def collectionClassName: String }].collectionClassName
   }
 }
```

kpodsiad:
I spoke with Wojciech and there should be the Scala Native release at the end of this week. That release will contain fixes for problems spotted here, so I'll wait for it.

armanbilge:
> That release will contain fixes for problems spotted here, so I'll wait for it.

Do you mean for `CrossVersion.for3Use2_13`? I must strongly recommend against using this technique.

When a widely-used library (such as scalacheck) is depended on with `CrossVersion.for3Use2_13` by another widely-used library (such as munit) this creates a binary-breaking division in the ecosystem: even after scalacheck is properly published for Native/Scala 3, this artifact is incompatible with the Native/2.13 artifact that munit is using, and thus the ecosystem will be split on this dependency.

For an example of how this plays out, I strongly recommend you read this thread.
- https://github.com/scala-js/scala-js-dom/issues/451

 **tl;dr** scala-js-dom and all its downstream users (very nearly the entire Scala.js ecosystem) had to do a breaking version jump.

Further reading:
- https://gist.github.com/djspiewak/b8f2b4547951442102488964bc351cf9#5-dont-rely-on-withdottycompat
  Note that `withDottyCompat` became `for3Use2_13`

kpodsiad:
@armanbilge thanks for this information. Just to confirm, the preferable solution is waiting until scalacheck is published properly for scala-native/scala3? In every other approach, munit ends up depending on the `_2.13` artifact.

SethTisue:
I just merged some PRs in the ScalaCheck repo that may help get things moving again.

armanbilge:
@kpodsiad how does would munit feel about scalacheck bumping to 3.1 on all platforms? See discussion in https://github.com/typelevel/scalacheck/pull/868#issuecomment-1067995736.

kpodsiad:
@armanbilge I'm afraid that I'm not the right person to ask.

@olafurpg @gabro could you share your opinion as creators and maintainers?

gabro:
IIUC this would only affect MUnit + Scalacheck on Scala Native.

If that's the case then I think it's fine (I'm assuming we can keep using the current version of Scalacheck for jvm/js targets in order to preserve compatibility with 3.1)

The only drawback is that downstream libraries targeting e.g. jvm and native would need to do the same (either depend on different versions of MUnit or break compatibility with 3.0)

I'll let @olafurpg call the shots, but it would be a terrible thing in my opinion

olafurpg:
No objections from me to drop 3.0 support

SethTisue:
> I'm assuming we can keep using the current version of Scalacheck for jvm/js targets in order to preserve compatibility with 3.0

one would assume so (I did too), but the assumption was called into question by @armanbilge at https://github.com/typelevel/scalacheck/pull/868#issuecomment-1067989391 , where he wrote:

> these sorts of hacks seem to cause problems see e.g. https://github.com/portable-scala/sbt-crossproject/issues/130

how strong an objection that is, I'm not sure — if we run into trouble, would it be minor trouble we could work around, or really serious trouble?

armanbilge:
Hmm, there seems to be some confusion (at least on my end :)

I was commenting on the potential difficulty of hacking a `build.sbt` so that some platforms publish on 3.0 and others on 3.1.

My understanding of @gabro's comment was that munit could continue to use the _current_ version of Scalacheck 1.15.4 on JVM/JS so that munit doesn't have to bump.

SethTisue:
it's difficult to speak precisely about this and difficult to be sure if understandings are shared

I guess I'll just say that I hope a simple-to-understand, simple-to-communicate-to-downstream-users solution turns out to be feasible here

armanbilge:
I agree. The other feasible solution seems to be waiting for the `-Yscala-release` flag (which now has a new name) and its sbt integration. But it's unclear to me whether/how this will work with Scala Native and what build hacks will be needed to accomplish that. Everyone moving to 3.1 is clean and easy.

gabro:
Given that we're early stage and we're all figuring this out, I sympathize with Seth's comment in the Scalacheck issue: it's probably for the best to just drop 3.0 and pretend it never happened

SethTisue:
perhaps @romanowski would like to try to talk us down from that radical position :-)

armanbilge:
No one's talked me down, so let's escalate 😉 
- https://github.com/typelevel/scalacheck/pull/879

armanbilge:
Scalacheck has now published a snapshot with Scala 3 / native support! I expect a stable release is imminent.

https://s01.oss.sonatype.org/content/repositories/snapshots/org/scalacheck/scalacheck_native0.4_3/1.16-0ac8005-SNAPSHOT/

armanbilge:
Your move! 😁 

https://repo1.maven.org/maven2/org/scalacheck/scalacheck_native0.4_3/

kpodsiad:
@armanbilge Could you help me with JS tests? They suddenly started failing after I bumped scala-js to the `1.9.0`. Although there is a clue what to do, I don't why it's needed at all and that's why I would appreciate help from someone with _any_ scala.js experience ;)

sjrd:
You're running into the [fixes with compatibility concerns](https://www.scala-js.org/news/2021/08/04/announcing-scalajs-1.7.0/#fixes-with-compatibility-concerns) about regular expressions in Scala.js 1.7.0. There are suggested strategies on how to avoid `MULTILINE` in [the regular expressions doc page of Scala.js](https://www.scala-js.org/doc/regular-expressions.html).

This is not my repo, but I would humbly suggest upgrading Scala.js in a separate PR to be able to address these changes independently of the larger changes in this PR.

sjrd:
It seems that the problematic regular expressions is at
https://github.com/scalameta/munit/blob/36aeb753b2f9b72636e33331a6825d7bcedb87a8/tests/shared/src/test/scala/munit/BaseFrameworkSuite.scala#L85-L86
This regexp doesn't even use `^` nor `$`, and is only used with `replaceAll` (so not any anchored situation). The `MULTILINE` flag has no effect in that case, so it can safely be removed without any other change.

armanbilge:
@kpodsiad does Seb's suggestion fix it (thank you!)? Otherwise I'll try and looking into it a bit later, I would _really_ like to unblock this :)

kpodsiad:
@armanbilge I'll try to look at this tomorrow or, more probably, next week. 

Because of `with Scala.js 1.8+` in
`Scalacheck 1.16.0 is binary compatible with the 1.15.x and 1.14.x series. It is published for Scala 2.12, 2.13, and 3.1+ with
Scala.js 1.8+ and Scala Native 0.4. This release is the first to support Scala 3 on the Native platform.`

it seems like I need to either:
- leave the old version of scalacheck for JS and use the newly published `1.16.0` for Native and JVM.
- just add the linkerConfig for tests (`scalaJSLinkerConfig ~= { _.withESFeatures(_.withESVersion(ESVersion.ES2018)) }`) because it's needed only in tests and it won't affect library users I think.

armanbilge:
Hum, if I understand https://github.com/scalameta/munit/pull/477#issuecomment-1092707746 correctly you just need to remove that `MULTILINE` and it should work.

Ok, I'll give it a try.

armanbilge:
Woops, I also see that this PR is targeting `main` instead of a `series/0.7` branch. If these changes are released as a 1.0 milestone, then most of the ecosystem won't be able to upgrade.

armanbilge:
Ok, put up a PR at:
- https://github.com/scalameta/munit/pull/511

armanbilge:
Hum, the multiline flag _is_ needed. The suggestion to bump the ES version for tests-only would work without affecting downstreams, but it would prevent your tests from being able to catch uses of ES2018 features in the published code. Which is not so great.
```scala
scalaJSLinkerConfig ~= { _.withESFeatures(_.withESVersion(ESVersion.ES2018)) }
```

kpodsiad:
@olafurpg approval will be gladly welcomed to fulfill formalities ;)

armanbilge:
Woops, fyi we've got conflicts now.

valencik:
Sorry, merge conflicts are my fault. I'll resolve those and then merge once CI passes.

valencik (approved):
After some discussion in Discord yesterday we're going to go ahead with this targeting 1.0 and soon cut a stable v1.0

These changes look reasonable to me, and CI is a nice shade of green :)

PR #478, opened 2022-01-06
Make flaky tag work with Scalacheck suites

olafurpg:
Previously, the Scalacheck suite transform was registered after the
`.flaky` transform so a flaky test would fail even if it was OK to
ignore flaky failures. Reordering the transforms fixes the issue.

### DIFF
--- a/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
+++ b/munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
@@ -10,7 +10,6 @@ import scala.util.Try
 import munit.internal.FutureCompat._
 
 trait ScalaCheckSuite extends FunSuite {
-
   def property(
       name: String
   )(body: => Prop)(implicit loc: Location): Unit = {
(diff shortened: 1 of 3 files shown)


munit-scalacheck/shared/src/main/scala/munit/ScalaCheckSuite.scala
```diff
@@ -29,7 +29,7 @@ trait ScalaCheckSuite extends FunSuite {
   implicit def unitToProp(unit: Unit): Prop = Prop.passed
 
   override def munitTestTransforms: List[TestTransform] =
-    super.munitTestTransforms :+ scalaCheckPropTransform
+    scalaCheckPropTransform +: super.munitTestTransforms
```
olafurpg, line 32: Feels like the discussions around transform registrations paid off big time. I was a big proponent for side-effecting transforms but that would have been a nightmare to debug for this bug here.

gabro (approved):
LGTM🙏

gabro:
Mm the CI failure looks legit

olafurpg:
The test failure was a legitimate bug, which only got surfaced now because of the change. The problem was that the JS/Native test reporter reported that ignored tests also succeeded.

PR #484, opened 2022-01-17
Try to reproduce #478

olafurpg:


tests/shared/src/main/scala/munit/Issue478FrameworkSuite.scala
```diff
@@ -0,0 +1,13 @@
+package munit
+
+class Issue478FrameworkSuite extends FunSuite {
+  override def munitFlakyOK: Boolean = true
+  test("boom".flaky.fail) {}
```
olafurpg, line 5: @danicheg this test is marked with `.fail.flaky` and the body always succeeds. My expectation is that this test gets treated as flaky since it succeeds despite the `.fail` flag.

danicheg: It's an uh-oh moment. Now I see the problem (we enable flaky tests only on the CI and not at the local). Sorry for the interruption.

PR #488, opened 2022-01-25
Scala Native and Scala 3

sideeffffect:


gabro:
Hi @sideeffffect thanks for the contribution! Have you checked out https://github.com/scalameta/munit/pull/477 already?

Unless I'm missing something this is doing the same thing?

sideeffffect:
Yup, the same goal. How can I help to move it forward? Can we turn it from draft to a real PR, run tests and merge if green?

gabro:
@kpodsiad is still working on it, so he can answer better, but I see there's only a failing test there so maybe there's not too much work left?

SethTisue:
curious about the status of this. do we need to find a new volunteer to finish it?

sideeffffect:
I think this PR is basically deprecated by https://github.com/scalameta/munit/pull/477

PR #501, opened 2022-03-12
feat: take into account TaskDefs with only TestSelectors.

kpodsiad:
`TestSelector` can be used to create `TaskDef` which should only execute selected tests from the given test class.

More info can be found at https://github.com/sbt/junit-interface/pull/108 (the changes are almost the same)

~Could you point me how to add a test for that feature and for compatibility?~
That was a silly question, I've already done tests for this in Junit, why not use them here as well.

tl;dr why
This will allow Metals to run a single munit test from the whole suite.

### DIFF
--- a/junit-interface/src/main/java/munit/internal/junitinterface/JUnitRunner.java
+++ b/junit-interface/src/main/java/munit/internal/junitinterface/JUnitRunner.java
@@ -11,7 +11,12 @@ import org.junit.runner.notification.RunListener;
 
 import sbt.testing.Runner;
 import sbt.testing.Task;
+import sbt.testing.Selector;
+import sbt.testing.TestSelector;
 import sbt.testing.TaskDef;
+
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 final class JUnitRunner implements Runner {
   private final String[] args;
@@ -127,7 +132,8 @@ final class JUnitRunner implements Runner {
     for (int i = 0; i < length; i++) {
       TaskDef taskDef = taskDefs[i];
       if (shouldRun(computer, taskDef)) {
-        tasks.add(new JUnitTask(this, settings, taskDef, computer));
+        RunSettings alteredSettings = alterRunSettings(this.settings, taskDef.selectors());
+        tasks.add(new JUnitTask(this, alteredSettings, taskDef, computer));
       }
     }
     return tasks.toArray(new Task[0]);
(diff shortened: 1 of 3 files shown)


tests/jvm/src/test/scala/munit/TestSelectorSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+ */
+class TestSelectorSuite extends FunSuite {
+  val framework = new Framework();
+  val runner: Runner = framework.runner(
+    Array.empty,
+    Array.empty,
+    this.getClass().getClassLoader()
+  );
+
+  val fingerprint = framework.munitFingerprint
```
kpodsiad, line 39: So this test mimics the behavior of sbt's `ForkMain` - it creates a framework instance which later will be used for running tests.

tests/jvm/src/test/scala/munit/TestSelectorSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+  private def getEventHandler(): (mutable.ListBuffer[String], EventHandler) = {
+    val executedItems = new mutable.ListBuffer[String]
+    val eventHandler = new EventHandler {
+      override def handle(event: Event) =
+        if (event.status() == Status.Success) {
+          executedItems += event.fullyQualifiedName()
+        }
+    }
+    (executedItems, eventHandler)
+  }
```
kpodsiad, line 54: Each executed test will send an event, so I use that fact to gather information about executed tests.

junit-interface/src/main/java/munit/internal/junitinterface/JUnitRunner.java
```diff
@@ -153,6 +158,30 @@ private RunListener createRunListener(String runListenerClassName) {
...
+   * from these selectors. This allows to run particular test cases within given test class.
+   * testFilter is treated as a regular expression, hence joining is done via '|'.
+   */
+  private RunSettings alterRunSettings(RunSettings defaultSettings, Selector[] selectors) {
+    boolean onlyTestSelectors =
+        Arrays.stream(selectors).allMatch(selector -> selector instanceof TestSelector);
+    if (onlyTestSelectors) {
+      String testFilter =
+          Arrays.stream(selectors)
+              .map(selector -> ((TestSelector) selector).testName())
```
olafurpg, line 173: Should we guard this cast?

olafurpg (approved):
LGTM! It’s very nice to have this functionality

junit-interface/src/main/java/munit/internal/junitinterface/JUnitRunner.java
kpodsiad: I think I can try to get rid off this cast

junit-interface/src/main/java/munit/internal/junitinterface/JUnitRunner.java
```diff
@@ -153,6 +159,37 @@ private RunListener createRunListener(String runListenerClassName) {
...
+        Arrays.stream(selectors)
+            .flatMap(
+                selector -> {
+                  if (selector instanceof TestSelector) {
+                    return Stream.of((TestSelector) selector);
+                  } else {
+                    return Stream.empty();
+                  }
+                })
+            .toList();
```
kpodsiad, line 179: @olafurpg this is the best what can I do in Java 😅 At least now, casting is done just after `instanceof` check

olafurpg: LGTM 👍

olafurpg: I think you can replace the if/else with a ternary operator to drop the block expression with return statements

kpodsiad:
I've applied suggestions and squashed commits - the PR should be mergeable now.
Could you publish the new version with these changes after merging @olafurpg?

olafurpg:
@kpodsiad As a member of the Scalameta org you have merge rights. Feel free to merge and push a git tag v1.0.0-M3 to trigger a release. I use the following scripts locally so I can execute `git release VERSION`
```
❯ cat `which git-release.sh`
set -eux
TAG="v$1"
echo $TAG
git tag -fa "$TAG" -m $TAG && git push -f origin $TAG

lsif-typescript on 🌱 cross-project via 🐳 default is 📦 v0.1.14 via ⬢ v16.7.0 on ☁️ olafurpg@sourcegraph.com
❯ cat ~/.gitconfig | grep git-release
	release = !git-release.sh $2
❯ git release 1.0.0-M3
```

kpodsiad:
I didn't want to abuse my membership in the project not related directly to Metals ;)

PR #521, opened 2022-04-15
Revive "Strict Equality" for `assertEquals()`

### LINKED ISSUE #189: Wrong order of arguments for assertEquals

valencik:
This PR is mostly a freshly rebased version of https://github.com/scalameta/munit/pull/225.

It is absolutely worth reviewing the discussion in the original PR. However, I've copy and pasted its description here:

> Previously, MUnit had a subtyping constraint on `assertEquals(a, b)` so
> that it would fail to compile if `a` was not a subtype of `b`. This was
> a suboptimal solution because the compile error messages could become
> cryptic in some cases. Additionally, this API didn't integrate with
> other libaries like Cats that has its own `cats.Eq[A,B]` type-class.
> 
> Now, MUnit uses a new `munit.Compare[A,B]` type-class for comparing
> values of different types. By default, MUnit provides a "universal"
> instance that permits comparison between all types and uses the built-in
> `==` method. Users can optionally enable "strict equality" by adding the
> compiler option `"-Xmacro-settings.munit.strictEquality"` in Scala 2.
> In Scala 3, we use the `Eql[A, B]` type-classes instead to determine
> type equality (per http://dotty.epfl.ch/docs/reference/contextual/multiversal-equality.html)
> 
> Note: the first few commits are unrelated to this feature, I cleaned up
> a few small quirks I noticed while working on this feature.
> 
> Fixes #189

### Binary Incompatibility

This PR introduces a binary breaking changes to the following methods:
- `munit.Assertions.assertNotEquals`
- `munit.Assertions.assertEquals`
- `munit.FunSuite.assertNotEquals`
- `munit.FunSuite.assertEquals`

### Rebasing notes:

The rebase got a little hairy. During it I ditched special casing for array comparisons, so 1503667ee29e28ce41a9649050c4816f6470cbe8 reintroduces it. Similarly, b28073f5e265e2a4ac59b4149548ba815ae961f8 reintroduces the better string inequality error messages.


munit/shared/src/main/scala/munit/Clue.scala
```diff
@@ -10,5 +10,7 @@ class Clue[+T](
   override def toString(): String = s"Clue($source, $value)"
 }
 object Clue extends MacroCompat.ClueMacro {
-  def empty[T](value: T): Clue[T] = new Clue("", value, "")
+  @deprecated("use fromValue instead", "0.8.0")
```
olafurpg, line 13: ?
```suggestion
  @deprecated("use fromValue instead", "1.0.0")
```

tests/shared/src/main/scala/munit/StackTraceFrameworkSuite.scala
```diff
@@ -28,9 +28,6 @@ object FullStackTraceFrameworkSuite
     extends BaseStackTraceFrameworkSuite(
       Array("-F"),
       """|at munit.Assertions:failComparison
-         |  at munit.Assertions:failComparison$
```
olafurpg, line 31: Nice 👍

tests/shared/src/test/scala/munit/AssertionsSuite.scala
```diff
@@ -50,12 +50,133 @@ class AssertionsSuite extends BaseSuite {
...
+           |
+           |    munit.Compare.compareSubtypeWithSupertype[List[Int], Vector[Int]](
+           |      /* missing */summon[List[Int] <:< Vector[Int]]
+           |    )
+           |
+           |But no implicit values were found that match type List[Int] <:< Vector[Int].
+           |
+           |The following import might make progress towards fixing the problem:
+           |
+           |  import munit.CustomCompare.fromCustomEquality
```
olafurpg, line 78: It's a shame these suggestions have false positives like this 🙈

tests/shared/src/test/scala/munit/AssertionsSuite.scala
```diff
@@ -50,12 +50,133 @@ class AssertionsSuite extends BaseSuite {
...
+           |Possible ways to fix this error:
+           |  Alternative 1: provide an implicit instance for Compare[A, B]
+           |  Alternative 2: upcast either type into `Any` or a shared supertype
+           |assertEquals(new A, new B)
+           |            ^
+           |""".stripMargin
+    )
+  }
+
+  test("chat-int-nok") {
```
olafurpg, line 143:
```suggestion
  test("char-int-nok") {
```

olafurpg (approved):
This was a lot of work! Thank you so much for reviving this PR ❤️ LGTM.

Only a few minor suggestions, feel free to ignore them

tests/shared/src/test/scala/munit/AssertionsSuite.scala
valencik: Mmmm, yeah this is a bummer.
I think it's related to https://github.com/lampepfl/dotty/issues/9685

olafurpg: Nice find! I left a comment in that issue about the fact that we can reproduce without `inline` https://github.com/lampepfl/dotty/issues/9685#issuecomment-1100684424

Issue #492, opened 2022-02-08
partial error reporting using chromedriver


bblfish:
When I use chromedriver to run tests in the browser in my [httpSig](https://github.com/bblfish/httpSig/tree/24ebb1c2c6f330b95bc5b05cb10da8b66a1ff579) project, I don't get the full diff error results as I would when running the java tests.

```scala
> ciChrome
```
results in the following output:
![Screenshot 2022-02-08 at 09 22 41](https://user-images.githubusercontent.com/124506/152946781-8ac23e3e-9018-422d-abde-1934118a5ec8.png)

Which does not show the diffs.

bblfish:
@armanbilge told me this is related to [#247 Avoid relying on Node's fs when running in browser-based JSEnv](https://github.com/scalameta/munit/issues/247). Perhaps there is a way of making the code available via HTTP, to get the diffs?

armanbilge:
There's also a proposal to solve this in https://github.com/scalameta/munit/pull/441#issuecomment-945084300.

PR #529, opened 2022-05-27
Add support for JSDOM to JSIO

armanbilge:
The JSDOM JSEnv runs in a sandbox within Node.js, so it has access to the same APIs via a sandbox-breaking technique. See:
- https://github.com/jsdom/jsdom/issues/2729
- https://github.com/scala-js/scala-js-macrotask-executor/pull/17
- https://github.com/scoverage/scalac-scoverage-plugin/pull/456

### DIFF
--- a/munit/js/src/main/scala/munit/internal/io/JSIO.scala
+++ b/munit/js/src/main/scala/munit/internal/io/JSIO.scala
@@ -6,7 +6,11 @@ import scala.util.Try
 object JSIO {
 
   private def require(module: String): Option[js.Dynamic] = {
-    Try(js.Dynamic.global.require(module)).toOption
+    Try(js.Dynamic.global.require(module)) // Node.js
+      .orElse( // JSDOM
+        Try(js.Dynamic.global.Node.constructor("return require")()(module))
+      )
+      .toOption
   }
   val process: Option[js.Dynamic] = require("process")
   val path: Option[js.Dynamic] = require("path")


valencik:
How would someone test that this works?

armanbilge:
It's slightly annoying to setup, but we have to find a way to run the test suite in the JSDOM JSEnv in addition to Node.js. https://github.com/scoverage/scalac-scoverage-plugin/pull/456 is one example of how to do that.
https://github.com/scala-js/scala-js-env-jsdom-nodejs/

armanbilge:
There we go. Enjoy your 1000+ lines of lockfiles :P

armanbilge:
Confirmed that JSDOM fails in CI without my changes in https://github.com/scalameta/munit/pull/530.

valencik:
Oh wow, thanks for that.

Is the idea that you could now run your munit tests in a browser environment?
I admit, I'm a tiny bit confused what this enables. It seems like using munit and jsdom was previously possible: https://github.com/scala-js/scala-js-env-jsdom-nodejs/#usage---writing-a-test

But this is different in that munit itself could run within the jsdom environment?

armanbilge:
Yes that's right, it was already possible to use it as a test environment. But any file based features were not working (such as diffs and line numbers) because browsers don't have access to file system. Since JSDOM is a mere sandbox emulating the browser, we can break out to access the relevant APIs.

armanbilge:
See also:
- https://github.com/scalameta/munit/issues/492

armanbilge:
And also:
- https://github.com/scalameta/munit/issues/247

valencik (approved):
This is really cool. Tested locally and everything works swell!

Just noting that this only runs the tests in the JSDOM environment in our CI, there isn't a new project within sbt that runs the tests.

You can manually run them in a sbt shell by changing your `jsEnv`:
```
set testsJS / jsEnv := {new org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv}`
```

And then run the js tests with:
`testsJS / test`

olafurpg (approved):
LGTM it's great that we can support jsdom 👏

Issue #541, opened 2022-06-25
Simplify and expose munit's `rootCause` util


valencik:
Munit has a handy utility for unwrapping exceptions to provide better error messages.

[`rootCause`](https://github.com/scalameta/munit/blob/64fee6586a51ba1670872825f25a0d77837a3adb/munit/shared/src/main/scala/munit/MUnitRunner.scala#L318-L326):
```scala
  private def rootCause(x: Throwable): Throwable = x match {
    case _: InvocationTargetException | _: ExceptionInInitializerError |
        _: UndeclaredThrowableException | _: ExecutionException
        if x.getCause != null =>
      rootCause(x.getCause)
    case _ => x
  }
```

This was introduce in https://github.com/scalameta/munit/pull/277 where it was ultimately copied from [a Scala compiler util](https://github.com/scala/scala/blob/69ac276159dbffa6c0f0d486fc1ba1ca6f5ff318/src/compiler/scala/tools/nsc/util/Exceptional.scala#L19-L27).

1. I think we should expose `rootCause` under `private[munit]` so that third-party integrations like [scalacheck-effect can use it without copying](https://github.com/typelevel/scalacheck-effect/blob/f146a3422d0b4fbdad3c3f725a8f110cdcc03cdb/munit/shared/src/main/scala/munit/ScalaCheckEffectSuite.scala#L96-L103).

2. We should remove the handling of `InvocationTargetException` and `UndeclaredThrowableException`, which are not available on ScalaJS.
- Ultimately I don't believe these two exceptions need to be unwrapped to address the issue reported in https://github.com/scalameta/munit/issues/275 which only needed unwrapping of `ExecutionException`
- Similarly, they are not needed to address the same issue in scalacheck-effect https://github.com/typelevel/scalacheck-effect/pull/196
- These shims are [now internal to munit](https://github.com/scalameta/munit/issues/437) meaning downstream libraries would also need to shim them.

I think the Scala compiler util included `InvocationTargetException` and `UndeclaredThrowableException` because it was used in signal handling code that dealt with more than just Futures. I think you can see this in the PR that introduced `UndeclaredThrowableException` to this code: https://github.com/scala/scala/commit/1fcb86507062f9923465cdf72e6b70c3a1cfe96b#diff-6624212dec6b02c76a7ef6d0a388f0d0f6b9ad71b520020d493d5c4880d9db3fR120

PR #542, opened 2022-06-25
Expose rootCause util

valencik:
Resolves https://github.com/scalameta/munit/issues/541 by making `rootCause` publicly available with `Exceptions.rootCause` but keeps things cross-platform.

cc. @armanbilge

### DIFF
--- a/munit/js/src/main/scala/munit/internal/PlatformCompat.scala
+++ b/munit/js/src/main/scala/munit/internal/PlatformCompat.scala
@@ -75,8 +75,4 @@ object PlatformCompat {
   private var myClassLoader: ClassLoader = _
   def setThisClassLoader(loader: ClassLoader): Unit = myClassLoader = loader
   def getThisClassLoader: ClassLoader = myClassLoader
-
-  type InvocationTargetException = munit.internal.InvocationTargetException
-  type UndeclaredThrowableException =
-    munit.internal.UndeclaredThrowableException
 }
--- a/munit/js/src/main/scala/munit/internal/PlatformCompat.scala
+++ b/munit/js/src/main/scala/munit/internal/PlatformCompat.scala
@@ -75,4 +75,8 @@ object PlatformCompat {
   private var myClassLoader: ClassLoader = _
   def setThisClassLoader(loader: ClassLoader): Unit = myClassLoader = loader
   def getThisClassLoader: ClassLoader = myClassLoader
+
+  type InvocationTargetException = munit.internal.InvocationTargetException
+  type UndeclaredThrowableException =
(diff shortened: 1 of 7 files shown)


munit/js/src/main/scala/munit/internal/PlatformCompat.scala
```diff
@@ -75,8 +75,4 @@ object PlatformCompat {
   private var myClassLoader: ClassLoader = _
   def setThisClassLoader(loader: ClassLoader): Unit = myClassLoader = loader
   def getThisClassLoader: ClassLoader = myClassLoader
-
-  type InvocationTargetException = munit.internal.InvocationTargetException
-  type UndeclaredThrowableException =
-    munit.internal.UndeclaredThrowableException
```
armanbilge, line 81: I think we can also remove these type aliases from the corresponding JVM/Native files.

munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -465,4 +455,13 @@ object MUnitRunner {
       case nsme: NoSuchMethodException => false
     }
   }
+
+  // NOTE(olafur): these exceptions appear when we await on futures. We unwrap
+  // these exception in order to provide more helpful error messages.
+  private[munit] def rootCause(x: Throwable): Throwable = x match {
```
armanbilge, line 461: We should add a note here to preserve bincompat since this is intended to be used by integrations.

munit/js/src/main/scala/munit/internal/PlatformCompat.scala
valencik: Nice catch

olafurpg (changes requested):
Can we move this method to some separate object? If it’s intended to be consumed externally then I’d prefer to have it be public (not package private) and on an object so that it’s clearly a stateless method (doesn’t need to live inside a class). 


I’d also prefer to keep all the cases of the old implementation. I recall hitting on InvocationTargetException in a lot of cases but it might not have a test case to reproduce it.

olafurpg (commented):
My bad, I missed that it’s a static method. I’d prefer to move the method to a new object named “Exceptions” or something like that. MUnitRunner will no longer be needed for JUnit 5 if that upgrade ever happens.

olafurpg (approved):
Approving just to unblock merge in case I miss notifications

valencik:
@armanbilge would you mind taking another peak at this now that cross-compatibility is back in? :)

armanbilge (approved):
Looks good, the diff is quite small anyway :)

PR #560, opened 2022-07-15
scalafix 0.10.1

bjaglin:
Closes https://github.com/scalameta/munit/pull/550
Closes https://github.com/scalameta/munit/pull/537
Follows https://github.com/scalameta/munit/pull/513#issuecomment-1095413509

As of `sbt-scalafix` 0.10.0, scala version is no longer overriden by `scalafixEnable`, so running with `++2.12.10` would result in "type nowarn is not a member of package annotation" as `nowarn` was introduced in 2.12.13 and sbt macros generate code that uses it.

```
[error] /home/runner/work/munit/munit/munit-sbt/src/main/scala/munit/sbtmunit/MUnitPlugin.scala:42:21: type nowarn is not a member of package annotation
[error]     munitBucketName := Some("munit-test-reports"),
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,11 +41,11 @@ mimaPreviousArtifacts := Set.empty
 crossScalaVersions := List()
 addCommandAlias(
   "scalafixAll",
-  "; ++2.12.10 ; scalafixEnable ; all scalafix test:scalafix"
+  s"; ++$scala212 ; scalafixEnable ; all scalafix test:scalafix"
 )
 addCommandAlias(
   "scalafixCheckAll",
-  "; ++2.12.10 ;  scalafixEnable ; scalafix --check ; test:scalafix --check"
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -41,11 +41,11 @@ mimaPreviousArtifacts := Set.empty
 crossScalaVersions := List()
 addCommandAlias(
   "scalafixAll",
-  "; ++2.12.10 ; scalafixEnable ; all scalafix test:scalafix"
```
bjaglin, line 44: before https://github.com/scalacenter/sbt-scalafix/pull/292, the patch version was ignored as sbt-scalafix uses to set its own version ([2.13 or later, containing `nowarn`](https://github.com/scala/scala/commit/ec8f9de10ca9be1a3b9ec4b0c039a24455f7bad5))

valencik (approved):
Thanks for this :)

Issue #563, opened 2022-07-22
override val scalaCheckInitialSeed = ... leads to NullPointerException


FrancescoSerra:
Hi, as per the output in [this issue](https://github.com/http4s/http4s/issues/6538), I've been overriding the scalaCheckInitialSeed, but when running the test I got:
```
==> X org.http4s.PathSuite.initializationError  0.002s java.lang.NullPointerException: Cannot invoke "String.length()" because "s$1" is null
    at org.scalacheck.rng.Seed$.$anonfun$fromBase64$1(Seed.scala:188)
    at scala.util.Try$.apply(Try.scala:210)
    at org.scalacheck.rng.Seed$.fromBase64(Seed.scala:187)
    at munit.ScalaCheckEffectSuite.$anonfun$initialSeed$1(ScalaCheckEffectSuite.scala:42)
    at scala.Option.getOrElse(Option.scala:201)
    at munit.ScalaCheckEffectSuite.$init$(ScalaCheckEffectSuite.scala:42)

```
However, `override **def** scalaCheckInitialSeed =` worked just fine.

armanbilge:
`lazy val` can probably work here too. But honestly a `def` seems fine for a `String` constant.

PR #564, opened 2022-07-22
Use `def` to override `scalaCheckInitialSeed`

armanbilge:
Fixes https://github.com/scalameta/munit/issues/563


FrancescoSerra (approved):
Perfect!

valencik (approved):
This looks like purely a documentation and testing change? As in, all we're doing is making the best practice more clear?

armanbilge:
Yup. Basically a `def` is less likely to go wrong due to subtle initialization order issues.

PR #601, opened 2022-11-09
MUnitRunner: avoid invoking fireTestStarted if a test is skipped. 

rpalcolea:
This moves down `notifier.fireTestStarted(description)` to only apply when a test is NOT ignored. This will not also follow the `org.junit.runner.notification.RunNotifier` contract but also unblock people in Gradle to use their enterprise features 

I wonder if this could be added to a potential v1.0.0-M7

Here is a build scan that shows the tests working now in Gradle with this change https://scans.gradle.com/s/t2og6zyoc3nzc/tests/overview


rpalcolea:
Tagging @valencik and @armanbilge since it appears you are the most active contributors lately

marcphilipp (approved):

tgodzik (approved):
The change makes sense :+1:

PR #606, opened 2022-11-17
Use daemon thread instead of non-daemon thread which prevent JVM from exiting

### LINKED ISSUE #540: Non-termination regression with scala-cli starting with 1.0.0-M2
Hi,

I have tried using latest munit versions with scala-cli and found out that termination does not occur if all tests pass since 1.0.0-M2 (M1 works just fine, as all the 0.7.x versions). I have a repro repo here:

https://github.com/lbialy/munit-nonterm-issue

Just run `scala-cli test .` in this directory and it should hang after printing that all tests passed.

My scala-cli version is 0.1.8.

lwronski:
Fixes #540.
Based on comment https://github.com/scalameta/munit/issues/540#issuecomment-1318675863

### DIFF
--- a/munit/jvm/src/main/scala/munit/internal/PlatformCompat.scala
+++ b/munit/jvm/src/main/scala/munit/internal/PlatformCompat.scala
@@ -4,15 +4,30 @@ import scala.concurrent.Future
 import sbt.testing.Task
 import sbt.testing.EventHandler
 import sbt.testing.Logger
+
 import scala.concurrent.duration.Duration
-import java.util.concurrent.Executors
+import java.util.concurrent.{
+  Executors,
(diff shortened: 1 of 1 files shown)


valencik (approved):
Testing locally with a `munitJVM/publishLocal` and then the following in `test.sc`:

```scala
//> using scala "2.13"
//> using lib "org.scalameta::munit::1.0.0-M6+60-d2685fa2-SNAPSHOT"
class SanityTest extends munit.FunSuite {
  test("sanity") {
    assert(2 + 2 == 4)
  }
}
```

Ran with `scala-cli test test.sc` we see things properly terminate. Previous versions hang.

PR #629, opened 2023-02-21
Handle assume(false) correctly

### LINKED ISSUE #614: Skipped tests reported as passing
Hello, I'm using `assume` to skip tests dynamically based on what's described in the documentation here: https://scalameta.org/munit/docs/filtering.html#ignore-single-test-case-based-on-a-dynamic-conditions, however the tests that fail the assertion are reported as passing (even though nothing is run). I'm not sure if I've misunderstood the documentation, from reading through some issues it looks like sometimes the final test report includes a "skipped" and sometimes not. I would like it to report as "n tests skipped", or at least ignore them completely but not report them as passing. I use the IntelliJ test API but it also happens when running with sbt.

The test itself is [here](https://github.com/aherlihy/carac/blob/main/src/test/scala/all/TestThief.scala#L32-L34) but if I simply run `assume(false)` in either `test(...)` or in `beforeEach`, it doesn't run the test but reports it as passed.

mzuehlke:
`assume(false)` gets now handled in the same way as `test("description".ignore")`. Added tests to cover both ways to skip tests.

Fixes #614

### DIFF
--- a/junit-interface/src/main/java/munit/internal/junitinterface/EventDispatcher.java
+++ b/junit-interface/src/main/java/munit/internal/junitinterface/EventDispatcher.java
@@ -129,7 +129,7 @@ final class EventDispatcher extends RunListener {
   public void testIgnored(Description desc) {
     postIfFirst(
         desc,
-        new InfoEvent(desc, Status.Skipped) {
+        new InfoEvent(desc, Status.Ignored) {
           void logTo(RichLogger logger) {
             logger.warn(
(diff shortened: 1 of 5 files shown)


tgodzik:
Looks like js and native tests are failing, could you take a look? Seems like they're `ignored` rather than `skipped`

olafurpg (approved):
LGTM 👍🏻 Thank you for this contribution!

Issue #573, opened 2022-09-07
Reusable suite-local fixtures is allocated after suite's beforeAll()


kamilkloch:
The following code
```scala
class TestMUnitResource extends munit.FunSuite {
  val resource = new Fixture[Double]("Double resource") {
    private var d: Double = Double.NaN

    def apply() = d

    override def beforeAll(): Unit = {
      println("Acquiring resource")
      d = 4.0
    }

    override def afterAll(): Unit = {
      println(s"Releasing resource $d")
      d = Double.NaN
    }
  }

  override def munitFixtures = List(resource)

  test("test1") {
    println(s"Using resource [${resource()}]")
  }

  override def beforeAll(): Unit = {
    println(s"Suite beforeAll: ${resource()}")
  }

  override def afterAll(): Unit = {
    println(s"Suite afterAll: ${resource()}")
  }
}
```
produces
```
Suite beforeAll: NaN
Acquiring resource
Using resource [4.0]
Suite afterAll: 4.0
Releasing resource 4.0
```
It looks that suite-local fixture is allocated _after_ `beforeAll` and released _after_ `afterAll` . Is this the desired behavior?

armanbilge:
Is this matching the behavior described in the breaking changes for munit 1.x?
https://github.com/scalameta/munit/releases/tag/v1.0.0-M1

kamilkloch:
> Is this matching the behavior described in the breaking changes for munit 1.x? https://github.com/scalameta/munit/releases/tag/v1.0.0-M1

It does indeed.

kamilkloch:
Done.

kamilkloch:
> But yes, this is the desired behaviour in the v1.0 release line. 

What is the rationale of such change, if I may ask? 
Example against such behavior: I would like to use suite-local fixture to create a shared api client, login some users in `beforeAll`, logout the users in `afterAll`. Not possible, as `beforeAll` runs before the fixture is ready.

mzuehlke:
Looking at the "issue" I realized that the fixtures (including the suite-local fixture) are teared town in the exact same order they are initialized. 

Wouldn't it make  more sense to call the `afterAll` and `afterEach` function on the `reverse` order of the fixtures ? 
So that the last created resources are destroyed first ?

So that `FixtureOrderFrameworkSuite` would look like (for only 1 test):
```
beforeAll(ad-hoc)
beforeAll(a)
beforeAll(b)
beforeEach(ad-hoc, 1)
[...]

PR #635, opened 2023-03-10
Invoke `after*` methods in reverse order

mzuehlke:
In case of multiple fixtures the "cleanup" methods `afterEach` and `afterAll` are called in the reverse order of `beforeEach` and `beforeAll`.

I am proposing to change this order. We did a similar breaking change in https://github.com/scalameta/munit/releases/tag/v1.0.0-M1 and are still on the road to a first stable v1.0.0.

Inspired by discussion in #573


munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -221,7 +221,7 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
       beforeAll: BeforeAllResult
   ): Future[Unit] = {
     sequenceFutures[Boolean](
-      beforeAll.loadedFixtures.iterator.map(f =>
+      beforeAll.loadedFixtures.reverseIterator.map(f =>
```
tgodzik, line 224: Do we need reverse here also?

mzuehlke:
Closing after the discussion in https://github.com/scalameta/munit/issues/573#issuecomment-1465269356

Issue #637, opened 2023-03-22
[Feature request] allow to override default printer


wahtique:
Hi ! 

I am confronted with a case where the default diff printer for iterable is less than helpful. Specifically, I have a case class ( a product ) wrapping a `ByteVector` ( an iterable ). When comparing two instances, the resulting diff look like this  : 

![image](https://user-images.githubusercontent.com/55251330/226890889-3c5836b0-67a6-4a3f-a9e0-df4dc0873338.png)

which is not very helpful as I do not actually want to compar byte by byte. The root cause seem to be 
 - values are different => munit will try to print the diff https://github.com/scalameta/munit/blob/main/munit/shared/src/main/scala/munit/Assertions.scala#L309 by using `munit.internal.console.Printers`
 - it finds a case class which is a product and recursively try to print each field https://github.com/scalameta/munit/blob/main/munit/shared/src/main/scala/munit/internal/console/Printers.scala#L102
 - the `ByteVector` is an iterable => compare element wise https://github.com/scalameta/munit/blob/main/munit/shared/src/main/scala/munit/internal/console/Printers.scala#L83

Rough expected behaviour would be : 

- Given a case class wrapping an scodecs `ByteVector` https://github.com/scodec/scodec-bits/blob/main/core/shared/src/main/scala/scodec/bits/ByteVector.scala 

```scala
//> using scala "3.2.2"
//> using dep  "org.scodec::scodec-bits:1.1.37"

import scodec.bits.ByteVector

final case class Wrapped(bv: ByteVector) {
  val stringValue: String = s"0x${bv.toHex}"
  override def toString: String = stringValue
}
``` 

- When I compare two instances

```scala
//> using dep "org.scalameta::munit0.7.29"

import munit.FunSuite

class Example extends FunSuite {
  test("This will fail") {
    val a = Wrapped(ByteVector.fromValidHex("0x1")
    val b = Wrapped(ByteVector.fromValidHex("0x2")
    assertEquals(a, b)
  }
}
```

- Then the diff should be 

````
- Wrapped(0x1)
+ Wrapped(0x2)
````

Would it be possible to : 
 - use traits instead of an object for printers so that users could override them when needed ? 
 - use implicits for printers to allow more fine grained control over diffs for specific types ?

wahtique:
Nvm foud the custom printer param :)

Issue #480, opened 2022-01-12
assertEquals produces unhelpful diff for large values


mvalle:
The `assertEquals` function generally produces a useful diff. But if the size of the objects that are being compared is too large, it produces an output that doesn't say anything useful about the assertion failure. 

Example test:
```
test("assertEquals on long lists") {
  val a = (1 to 158).toList
  val b = (1 to 158).toList ++ List(159)
  assertEquals(a, b)
}
```
output:
```
=> Diff (- obtained, + expected)
   ...,
+  ...,
   ...
```

kschwarz1116:
I was able to override the limit in the following way:
```scala
package example

import munit.{FunSuite, Printer}
import munit.internal.console.Printers

class FunTest extends FunSuite {
  def longPrinter = new Printer {
    def print(value: Any, out: StringBuilder, indent: Int): Boolean = false
    override def height = 200
  }
  override def munitPrint(clue: => Any): String = {
    clue match {
      case message: String => message
      case value           => Printers.print(value, longPrinter)
    }
  }
  test("assertEquals on long lists") {
    val a = (1 to 158).toList
    val b = (1 to 158).toList ++ List(159)
    assertEquals(a, b)
  }
}
```

Probably wouldn't hurt to add documentation around this.

mzuehlke:
#640 has probably solved this issue and it can be closed

mzuehlke:
The mentioned PR got merged before M8

PR #640, opened 2023-03-27
feat(printer): mk printers more easily configurable

wahtique:
Hi ! 

Following this issue I opened https://github.com/scalameta/munit/issues/637 and closed right after once I realised I should havve looked around a bit more, I tried my hand at using custom pretty printers to clarify some of my diffs. 

Using the answer to this question https://github.com/scalameta/munit/issues/480 I got pretty much what a I needed ( comparing scodecs ByteVectors ) but I found the experience a bit confusing and felt some boilerplate could be avoided.

Which leads me to this PR :) 

Changes : 

- add `orElse` combinator for `Printer` for composability
- add utility constructor for easier instantiation : user will only pass whatever is actually meaningful to them in most cases, ie. the custom logic to turn their test values into differenciable strings
- update `Assertion` to allow users to pass their custom `Printer` without also overriding `munitPrint`

### DIFF
--- a/munit/shared/src/main/scala/munit/Printer.scala
+++ b/munit/shared/src/main/scala/munit/Printer.scala
@@ -14,6 +14,78 @@ trait Printer {
   def height: Int = 100
   def isMultiline(string: String): Boolean =
     string.contains('\n')
+
+  /**
+   * Combine two printers into a single printer.
+   *
+   * Order is important : this printer will be tried first, then the other printer.
+   * The new Printer's height will be the max of the two printers' heights.
+   * 
+   * Comibining two printers can be useful if you want to customize Printer for
+   * some types somewhere, but want to add more specialized printers for some tests 
+   * 
+   * {{{
+   * trait MySuites extends FunSuite { 
+   *   override val printer = Printer { 
+   *     case long => s"${l}L"
+   *   }
+   * }
+   * 
+   * trait SomeOtherSuites extends MySuites {
+   *   override val printer = Printer {
+   *     case m: SomeCaseClass => m.someCustomToString
+   *   } orElse super.printer
+   * }
+   * 
+   * }}}
+   */
+  def orElse(other: Printer): Printer =
+    new Printer {
+      def print(value: Any, out: StringBuilder, indent: Int): Boolean =
+        this.print(value, out, indent) || other.print(
+          value,
+          out,
+          indent
+        )
+      override def height: Int = this.height.max(other.height)
+    }
+
+  val a = Byte
+}
+
+object Printer {
+
+  def apply(h: Int)(partialPrint: PartialFunction[Any, String]): Printer =
+    new Printer {
+      def print(value: Any, out: StringBuilder, indent: Int): Boolean =
+        value match {
+          case simpleValue =>
+            partialPrint.lift.apply(simpleValue).fold(false) { string =>
+              out.append(string)
+              true
+            }
+        }
+
+      override def height: Int = h
+    }
+
+  /**
+   * Utiliy constructor defining a printer for some types.
+   *
+   * This might be useful for some types which default pretty-printers
+   * do not output helpful diffs.
+   *
+   * {{{
+   * type ByteArray = Array[Byte]
+   * val listPrinter = Printer {
+   *   case ll: ByteArray => ll.map(String.format("%02x", b)).mkString(" ")
+   * }
+   * val bytes = Array[Byte](1, 5, 8, 24)
+   * Printers.print(bytes, listPrinter) // "01 05 08 18"
+   * }}}
+   */
+  def apply(partialPrint: PartialFunction[Any, String]): Printer =
+    apply(100)(partialPrint)
 }
 
 /** Default printer that does not customize the pretty-printer */
--- a/munit/shared/src/main/scala/munit/Printer.scala
+++ b/munit/shared/src/main/scala/munit/Printer.scala
@@ -11,7 +11,7 @@ trait Printer {
    * Returns true if this value has been printed, false if FunSuite should fallback to the default pretty-printer.
    */
   def print(value: Any, out: StringBuilder, indent: Int): Boolean
-  def height: Int = 100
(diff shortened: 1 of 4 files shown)


munit/shared/src/main/scala/munit/Printer.scala
```diff
@@ -14,6 +14,78 @@ trait Printer {
...
+      def print(value: Any, out: StringBuilder, indent: Int): Boolean =
+        this.print(value, out, indent) || other.print(
+          value,
+          out,
+          indent
+        )
+      override def height: Int = this.height.max(other.height)
+    }
+
+  val a = Byte
```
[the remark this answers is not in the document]
wahtique: Oops indeed must have escaped my worksheet ^^

wahtique:
Comment taken into account. I also restricted some tests to scala 2.13 like in `munit.PrintersSuite` and fixed the `Printer` implem to compile under 2.12

tests/shared/src/test/scala/munit/CustomPrinterSuite.scala
```diff
@@ -0,0 +1,141 @@
...
+      assertEquals(obtained, expected)
+    }
+  }
+
+  checkProp("list and int in product", isEnabled = isScala213) {
+    forAll(genFooBar) { foobar =>
+      val obtained = Printers
+        .print(foobar, listPrinter.orElse(intPrinter))
+        .stripMargin
+        .stripLeading()
```
tgodzik, line 99: Looks like `.stripLeading()` was added in JDK 11, so this will not compile currently.

wahtique: image

Fixed ! This made the updated test a bit less specific but should be acceptable since more thorough checks on whitespaces, indents etc. are made elsewhere

tgodzik:
Looks like the tests are passing! Let's see if anyone more familiar with the codebase has time to look at it.

munit/shared/src/main/scala/munit/Printer.scala
```diff
@@ -14,6 +14,84 @@ trait Printer {
...
+   * type ByteArray = Array[Byte]
+   * val listPrinter = Printer.apply {
+   *   case ll: ByteArray => ll.map(String.format("%02x", b)).mkString(" ")
+   * }
+   * val bytes = Array[Byte](1, 5, 8, 24)
+   * Printers.print(bytes, listPrinter) // "01 05 08 18"
+   * }}}
+   */
+  def apply(partialPrint: PartialFunction[Any, String]): Printer =
+    apply(100)(partialPrint)
```
gabro, line 94: we're repeating the default `100` height here and in the trait. Could we maybe extract it into a `val` in the object and use it to initialize the trait?


Something like:

```scala
trait Printer { 
  def height = Printer.defaultHeight
}

object Printer {
  val defaultHeight = 100

	def apply(partialPrint: PartialFunction[Any, String]): Printer =
    apply(defaultHeight)(partialPrint)
}
```

munit/shared/src/main/scala/munit/Printer.scala
```diff
@@ -14,6 +14,84 @@ trait Printer {
...
+    val h = height
+    new Printer {
+      def print(value: Any, out: StringBuilder, indent: Int): Boolean =
+        value match {
+          case simpleValue =>
+            partialPrint.lift.apply(simpleValue).fold(false) { string =>
+              out.append(string)
+              true
+            }
+        }
```
gabro, line 72: doesn't this match expression always match? What's the difference with:

```scala
def print(value: Any, out: StringBuilder, indent: Int): Boolean =
  partialPrint.lift.apply(value).fold(false) { string =>
    out.append(string)
    true
  }
```

Also, super minor preference, but I find `fold` not super straightforward to read compared to:

```scala
def print(value: Any, out: StringBuilder, indent: Int): Boolean =
  partialPrint.lift.apply(value) match {
    case Some(string) =>
    	out.append(string)
    	true
    case None => false
  }
```

wahtique: Damnit yeah it always match Must be a leftover from when I tried to include some recursivity and handle Arrays, Products....before realising it was a repeat of the logic which should be applied after the custom printer. Will remove it. Same for the fold !

wahtique:
Comments taken into account in the latest commits. Also added section covering the why and how of custom printers in https://scalameta.org/munit/docs/tests.html right under custom timeouts. Please tell me if its clear enough and / or if you would prefer it elsewhere :)

munit/shared/src/main/scala/munit/Printer.scala
```diff
@@ -11,12 +11,98 @@ trait Printer {
...
   def isMultiline(string: String): Boolean =
     string.contains('\n')
+
+  /**
+   * Combine two printers into a single printer.
+   *
+   * Order is important : this printer will be tried first, then the other printer.
+   * The new Printer's height will be the max of the two printers' heights.
+   *
+   * Exampple use case :  define some default printers for some types for all tests,
```
tgodzik, line 24:
```suggestion
   * Example use case :  define some default printers for some types for all tests,
```

tgodzik (changes requested):
Could you run `scalafix` and `scalafmt` ? Otherwise looks good, unless @gabro sees anything more to fix?

gabro (approved):
Looking good to me :)

wahtique:
scalafmt and scalafix ran without apparent issue

But I've got some errors from `doc/docusaurusCreateSite` which seem unrelated to this PR 
image

Any idea why ?

tgodzik (approved):
LGTM!

tgodzik:
> scalafmt and scalafix ran without apparent issue
> 
> But I've got some errors from `doc/docusaurusCreateSite` which seem unrelated to this PR image
> 
> Any idea why ?

That might be an existing issue, I don't think your change could have influenced anything

alexgb1:
Any plans to release this change soon ?

PR #646, opened 2023-04-15
Use js-native folder by sbt-crossproject

lolgab:
Avoid creating extra `Compile / unmanagedSourceDirectories` for non-jvm


tgodzik (approved):
LGTM thanks!

PR #648, opened 2023-04-18
Handle a StackOverflowError in addition to NonFatal errors

### LINKED ISSUE #583: Exception thrown inside a test crashes the runner
Hi!

Running the following code (`munit.scala`):

```scala
//> using lib "junit:junit:4.11"

import org.junit.runner.notification.RunListener
import org.junit.runner.{Description, JUnitCore}

class FatalErrorSuite extends munit.FunSuite:
  test("simple test") { () }
  test("failed test") {
    def loop(x: Int): Int = loop(x) + loop(x)
    loop(0)
  }
  test("another test") { () }

@main def main = 
  val junitCore = JUnitCore()
  val pointsCounter = PointsCounter()
  junitCore.addListener(pointsCounter)
  val result = junitCore.run(classOf[FatalErrorSuite])
  println(pointsCounter.runCount)

class PointsCounter(var runCount: Int = 0) extends RunListener:
  override def testStarted(description: Description): Unit =
    runCount += 1
```

Prints `3` as I would expect with munit 0.7:

```
➜ scala-cli run -d "org.scalameta::munit::0.7.27" munit.scala  
Compiling project (Scala 3.1.3, JVM)
Compiled project (Scala 3.1.3, JVM)
3
```

But prints `2` with munit 1.0.0-M1:

```
➜ scala-cli run -d "org.scalameta::munit::1.0.0-M1" munit.scala
Compiling project (Scala 3.1.3, JVM)
Compiled project (Scala 3.1.3, JVM)
2
```

And prints `2` with munit 1.0.0-M2 and stays stuck:

```
➜ scala-cli run -d "org.scalameta::munit::1.0.0-M2" munit.scala
Compiling project (Scala 3.1.3, JVM)
Compiled project (Scala 3.1.3, JVM)
2
^C
```

Seems like a bug; is it?

mzuehlke:
if they are happening inside a test

I only execute the test on the JVM, because
- on JS the error message is different
- on Native the runner just terminates (I know too little about ScalaNative)

Fixes #583

### DIFF
--- a/munit/shared/src/main/scala/munit/MUnitRunner.scala
+++ b/munit/shared/src/main/scala/munit/MUnitRunner.scala
@@ -285,25 +285,31 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
     }
 
     notifier.fireTestStarted(description)
+    def handleNonFatalOrStackOverflow(ex: Throwable): Future[Unit] = {
+      trimStackTrace(ex)
+      val cause = Exceptions.rootCause(ex)
+      val failure = new Failure(description, cause)
+      cause match {
+        case _: AssumptionViolatedException =>
+          notifier.fireTestAssumptionFailed(failure)
+        case _: FailSuiteException =>
+          suiteAborted = true
+          notifier.fireTestFailure(failure)
+        case _ =>
+          notifier.fireTestFailure(failure)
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM! Thanks!

Issue #149, opened 2020-06-19
Support NO_COLOR environment variable


olafurpg:
Currently, the only way to disable ANSI color output from MUnit is to use the cryptic combination of the testing framework flags `testOnly -- -n +l`. We should make it easier to disable colors in MUnit by respecting the NO_COLOR environment variable https://no-color.org/

worace:
This would be great to have. In the process of looking for this I also learned that there is an existing `--no-color` command line flag accepted by SBT. This seems to work to disable color output for scalatest but not for MUnit.

Would it make sense for MUnit to honor that flag as well, or is it too sbt-specific?

pokle:
I'd like there to be a way to turn off all terminal escape codes - because they just create a lot of noise in the log output of my build tool (buildkite)

mzuehlke:
This got fixed by
*  #703

PR #651, opened 2023-04-25
feat: respect NO_COLOR env variable

susliko:
Resolves https://github.com/scalameta/munit/issues/149

This PR makes use of the [NO_COLOR](http://no-color.org/) environment variable, disabling all ansi escapes when `NO_COLOR=1`.

Example:
```scala
//> using scala "2.13.10"
//> using lib "org.scalameta::munit::1.0.0-SNAPSHOT" // published locally

  test("foo") {
    assert(2 + 2 == 4)
  }
}
```

### DIFF
--- a/munit/shared/src/main/scala/munit/internal/console/AnsiColors.scala
+++ b/munit/shared/src/main/scala/munit/internal/console/AnsiColors.scala
@@ -1,23 +1,28 @@
 package munit.internal.console
 
 object AnsiColors {
-  val LightRed = "\u001b[91m"
+  // Foreground colors
+  val BLUE = "\u001B[34m"
+  val CYAN = "\u001B[36m"
+  val DarkGrey = "\u001B[90m"
+  val GREEN = "\u001B[32m"
   val LightGreen = "\u001b[92m"
-  val Reset = "\u001b[0m"
-  val Reversed = "\u001b[7m"
-  val Bold = "\u001b[1m"
-  val Faint = "\u001b[2m"
+  val LightRed = "\u001b[91m"
+  val Magenta = "\u001B[35m"
   val RED = "\u001B[31m"
   val YELLOW = "\u001B[33m"
-  val BLUE = "\u001B[34m"
-  val Magenta = "\u001B[35m"
-  val CYAN = "\u001B[36m"
-  val GREEN = "\u001B[32m"
-  val DarkGrey = "\u001B[90m"
 
-  def c(s: String, colorSequence: String): String =
-    if (colorSequence == null) s
+  // Styles
+  val Bold = "\u001b[1m"
+  val Faint = "\u001b[2m"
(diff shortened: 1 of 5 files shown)


susliko:
@valencik Thank you for taking a look!

> As you can see in the [mima run](https://github.com/scalameta/munit/actions/runs/4803233393/jobs/8547826117?pr=651), this is causing some binary incompatibility issues.

Mima issues were solely caused by the removal of unused escapes from `AnsiColors`. This change is reverted now, CI should be passing

>  It's not clear to me why they changes to JUnitReporter.scala, Lines.scala, or Diff.scala are necessary.

These files had some string concatenations with direct inclusion of ansi codes. This PR refactors them to use `Ansi Colors.c` method instead. Please, let me know, if there is another way to ensure escape codes to be filtered while leaving these files untouched

munit/shared/src/main/scala/munit/internal/console/AnsiColors.scala
```diff
@@ -1,23 +1,28 @@
...
-    if (colorSequence == null) s
+  // Styles
+  val Bold = "\u001b[1m"
+  val Faint = "\u001b[2m"
+  val Reset = "\u001b[0m"
+  val Reversed = "\u001b[7m"
+
+  def c(s: String, colorSequence: String): String = {
+    val isNoColorEnvSet = System.getenv("NO_COLOR") == "1"
+    if (colorSequence == null || isNoColorEnvSet) s
```
kitbellew, line 23: why is this implementation different from Ansi above? please don't read the env var if colorSequence is null, just like above.

munit/shared/src/main/scala/munit/internal/difflib/Diff.scala
```diff
@@ -43,9 +43,9 @@ class Diff(val obtained: String, val expected: String) extends Serializable {
 
   private def appendDiffOnlyReport(sb: StringBuilder): Unit = {
     header("Diff", sb)
-    sb.append(
-      s" (${AnsiColors.LightRed}- obtained${AnsiColors.Reset}, ${AnsiColors.LightGreen}+ expected${AnsiColors.Reset})"
-    ).append("\n")
+    val obtained = AnsiColors.c("- obtained", AnsiColors.LightRed)
+    val expected = AnsiColors.c("+ expected", AnsiColors.LightGreen)
+    sb.append(s" ($obtained, $expected)").append("\n")
```
kitbellew, line 48: looks like you're changing output to use `.c` instead of explicit color manipulation. please do that in a separate PR.

kitbellew (changes requested):

valencik:
Thank you again for this contribution!
Closing in favour of https://github.com/scalameta/munit/pull/703 which was able to minimize the diff nicely.

PR #663, opened 2023-06-13
Remove import suggestions from assertions

julienrf:
Needs https://github.com/lampepfl/dotty/pull/17924 to be merged first.


julienrf:
The corresponding PR has been merged in the compiler, but it is not released yet. Probably the commit of this PR will be needed when you will bump the Scala version of the project.

Issue #650, opened 2023-04-21
Sometimes, errors and failures during a test aren't caught.


markehammons:
If you look at my repository [here](https://github.com/markehammons/slinc/commit/24099dcdf73b2b64f3d2bd971bd3245f3e9ffe0b) you'll see that my tests are all passing for this commit. However, only 3 tests from BindingSpec were being reported on. After noticing this and investigating, I found that the ZLib test I had written was crashing and stopping the evaluation of the tests following it. This has been a frustratingly common situation with Munit, and I'm not sure if it's because of the nature of my code, or something that needs resolved in Munit.

SethTisue:
Are you to construct a self-contained, minimal reproduction?

markehammons:
I will give it a shot.

szymon-rd:
@markehammons Did you manage to find some minimization?

markehammons:
@szymon-rd I haven't had time these past weeks to try. I will try to do that once I catch up.

FloWi:
Hi!

I got bitten by the same bug as in the linked issue. I reproduced it in an example project.

https://github.com/FloWi/munit-bug-experiment/blob/main/src/test/scala/example/HelloSpec.scala

In my original project I used `Either.catchNonFatal` wrong. Apparently it was a fatal exception that I tried to catch in an Either.

This is the output
```
before assertion in first test
before wrong use of Either.catchNonFatal
example.HelloSpec:
  + first test 0.005s
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[success] Total time: 1 s, completed 02.07.2023, 16:34:35
```

As you can see, the 1st test is displayed correctly. The 2nd is being executed (the println statement gets written to stdout) but _not_ reported as a failure. It's not displayed at all. The 3rd test doesn't even get executed.

I would prefer mUnit to show the crashed test correctly and not silently hide it.

FloWi:
It seems that the bug has been introduced in the v1 branch of munit. I tried it with these dependencies
```sbt
    libraryDependencies ++= Seq(
      "org.typelevel" %% "cats-core" % "2.9.0",
      "org.scalameta" %% "munit" % "0.7.29" % Test,
    )
``´
and it crashed as expected but reports the error.

```
before assertion in first test
before wrong use of Either.catchNonFatal
example.HelloSpec:
  + first test 0.003s
==> X example.HelloSpec.expected error running tests  0.001s java.lang.IllegalAccessError: asdf
[...]

PR #669, opened 2023-07-06
In case of any Throwable inside a test mark the test as failed.

mzuehlke:
closes: #650


munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -310,6 +310,9 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
         handleNonFatalOrStackOverflow(ex)
       case ex: StackOverflowError =>
         handleNonFatalOrStackOverflow(ex)
+      case ex =>
```
filipwiech, line 313: If this is a fatal exception, would it make sense to abort the suite then, or not really/it doesn't matter? :thinking:

mzuehlke: Very good idea 👍

mzuehlke:
@tgodzik  do you have an opinion about this pull request.

tgodzik (approved):
Look good to me though I haven't touched the codebase too much

PR #677, opened 2023-07-19
Remove outdated doc instruction

Daenyth:
I'm told this step isn't needed any more


mzuehlke:
That is no longer needed since sbt 1.5.0

valencik (approved):
I think this is safe to remove. In any event, the actual quick-start page we're linking to also includes the "sbt lower than 1.5.0" caveat.

mzuehlke (approved):

Issue #679, opened 2023-07-27
`intercept[AssertionError]` should fail if no `AssertionError` is thrown


aymanelamyaghri:
This test should fail but it succeeds:

```scala
package example

class MyTests extends munit.FunSuite {
    test("intercept AssertionError") {
        intercept[AssertionError](println("Hello"))
    }
}
```

The output is:
```
Hello
example.MyTests:
  + intercept AssertionError 0.168s
```

PR #683, opened 2023-08-01
fix intercept for AssertionError

mzuehlke:
When trying to `intercept` an `AssertionError` till now the test passed successfully. 
This fix treats the FailException special.

closes: #679

### DIFF
--- a/munit/shared/src/main/scala/munit/Assertions.scala
+++ b/munit/shared/src/main/scala/munit/Assertions.scala
@@ -189,14 +189,15 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
       expectedExceptionMessage: Option[String],
       body: => Any
   )(implicit T: ClassTag[T], loc: Location): T = {
+    val expectedExceptionMsg =
+      s"expected exception of type '${T.runtimeClass.getName()}' but body evaluated successfully"
     try {
       body
-      fail(
-        s"expected exception of type '${T.runtimeClass.getName()}' but body evaluated successfully"
-      )
+      fail(expectedExceptionMsg)
     } catch {
       case e: FailExceptionLike[_]
-          if !T.runtimeClass.isAssignableFrom(e.getClass()) =>
+          if !T.runtimeClass.isAssignableFrom(e.getClass()) ||
+            e.getMessage.contains(expectedExceptionMsg) =>
         throw e
       case NonFatal(e) =>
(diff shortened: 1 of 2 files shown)


mzuehlke:
Is there a reason that this pull request hasn't been merged, yet ?

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -189,14 +189,15 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
...
       body
-      fail(
-        s"expected exception of type '${T.runtimeClass.getName()}' but body evaluated successfully"
-      )
+      fail(expectedExceptionMsg)
     } catch {
       case e: FailExceptionLike[_]
-          if !T.runtimeClass.isAssignableFrom(e.getClass()) =>
+          if !T.runtimeClass.isAssignableFrom(e.getClass()) ||
+            e.getMessage.contains(expectedExceptionMsg) =>
```
tgodzik, line 200: maybe a separate case:
```
```suggestion
           case failed: FailException if  failed.getMessage() == expectedExceptionMsg  =>
             throw e
```
to be super sure? So that we only employ this logic if someone throws exactly the same error as we would. Then they're on their own

tgodzik (changes requested):
Sorry! I wasn't super sure about the changes and was hoping for someone more knowledgeable to come by. I double checked and rethought the PR to offer some suggestions.

munit/shared/src/main/scala/munit/Assertions.scala
weipingc: I was caught by this AssertionError bug today, and found this fix. @tgodzik 's suggestion is good, but 
```
failed.getMessage() == expectedExceptionMsg
```
need to be changed to
```
failed.getMessage.contains(expectedExceptionMsg)
```
Tested this and @mzuehlke 's original change, both worked.

Can this be merged?

tgodzik: I totally forgot about this, @mzuehlke what do you think about suggested changes? I am ok also merging as is, we could update the test case though.

mzuehlke: I had forgotten about it, too. Sorry for that.
I adopted your suggestion.

tgodzik (approved):
LGTM!

Issue #692, opened 2023-09-05
Async tests failing on Scala Native


JD557:
Async tests on Scala Native fail with an unexpected opcode 4.

There's more info about the issue here: https://github.com/scala-native/scala-native/issues/3464#issuecomment-1705503874

That comment includes a possible fix on munit's side. I think the alternative would be to fix this once Scala Native 0.5.0 lands

PR #693, opened 2023-09-05
Call `scalanative.runtime.loop()`

armanbilge:
I'm surprised that this was missing and nobody noticed for so long. Every other test framework does this, it's absolutely essential.

Fixes https://github.com/scalameta/munit/issues/692.

### DIFF
--- a/munit/native/src/main/scala/munit/internal/PlatformCompat.scala
+++ b/munit/native/src/main/scala/munit/internal/PlatformCompat.scala
@@ -24,7 +24,9 @@ object PlatformCompat {
       duration: Duration,
       ec: ExecutionContext
   ): Future[T] = {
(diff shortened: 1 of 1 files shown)


armanbilge:
For example:

https://github.com/scala-native/scala-native/blob/aa9358858ce4bf9da95b9c78eadc5ab55f483578/junit-async/native/src/main/scala/scala/scalanative/junit/async/package.scala#L7-L10

https://github.com/etorreborre/specs2/blob/b4fb0a16f1a0369fe69eb994220317a5aaf2f3d9/common/native/src/main/scala/org/specs2/concurrent/package.scala#L8-L11

tgodzik (approved):
Thanks 🙏

Issue #695, opened 2023-09-14
Broken Scala Native tests on v1.0.0-M9


valencik:
The latest milestone release ([v1.0.0-M9](https://github.com/scalameta/munit/releases/tag/v1.0.0-M9)) appears to break tests on Scala Native.
The likely culprit is some complication with https://github.com/scalameta/munit/pull/693 interacting with cats-effect.

Related links:

- [munit-cats-effect failing build on M9](https://github.com/typelevel/munit-cats-effect/actions/runs/6188155190/job/16799550327)

- https://github.com/scala-native/scala-native/issues/3464

**Edit**: If downstream users see an error like:

```
[error] /usr/bin/ld: /home/runner/work/lucille/lucille/core/.native/target/scala-3.3.1/native-test/native-code-nativelib_native0.4_3-0.4.15-1/scala-native/gc/immix/Marker.c.o: in function `Object_IsWeakReference':
[error] Marker.c:(.text+0x88): undefined reference to `__weak_ref_ids_min'
[error] /usr/bin/ld: Marker.c:(.text+0xa2): undefined reference to `__weak_ref_ids_max'
[error] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This is actually caused by not being on the latest [Scala Native 0.4.15](https://github.com/scala-native/scala-native/releases/tag/v0.4.15)

armanbilge:
More details in this CE issue.

- https://github.com/typelevel/cats-effect/issues/3827

I'm operating under the assumption that this is a CE bug, but I'm not sure yet.

armanbilge:
I was able to reproduce without CE.

```scala
//> using dep org.scalameta::munit::1.0.0-M9
//> using nativeVersion 0.4.15

import scala.concurrent.duration.*
import scala.concurrent.*

class MySuite extends munit.FunSuite:
  override implicit def munitExecutionContext = ExecutionContext.global

  test("boo"):
    val p = Promise[Unit]()
    ExecutionContext.global.execute { () =>
      Thread.sleep(1000)
      p.success(())
    }
    p.future
```

valencik:
One tiny step further, we can drop the `Thread.sleep()` and still get the failure on native.
Note that this now also runs on Scala JS and passes there.

```scala
//> using dep org.scalameta::munit::1.0.0-M9
//> using nativeVersion 0.4.15

import scala.concurrent.duration.*
import scala.concurrent.*

class MySuite extends munit.FunSuite:
[...]

PR #700, opened 2023-09-15
Move `scalanative.runtime.loop()` invocation to correct place

armanbilge:
https://github.com/scalameta/munit/issues/693 had the right idea, but it was invoking `scalanative.runtime.loop()` in the wrong place. We should use it to `Await.result` on a `Future`.

Resolves https://github.com/scalameta/munit/issues/695


munit/native/src/main/scala/munit/internal/junitinterface/JUnitTask.scala
```diff
@@ -0,0 +1,43 @@
...
+      loggers: Array[Logger]
+  ): Array[Task] = {
+    PlatformCompat.newRunner(taskDef(), classLoader) match {
+      case None =>
+      case Some(runner) =>
+        runner.filter(runSettings.tags)
+        val reporter =
+          new JUnitReporter(eventHandler, loggers, runSettings, taskDef())
+        val notifier: RunNotifier = new MUnitRunNotifier(reporter)
+        runner.run(notifier)
```
armanbilge, line 38: JS uses `runAsync` here but actually we want to use `run` (sync) on Native.

munit/native/src/main/scala/munit/internal/PlatformCompat.scala
```diff
@@ -7,10 +7,17 @@ import scala.scalanative.reflect.Reflect
...
+import scala.concurrent.Await
+import scala.concurrent.Awaitable
 import scala.concurrent.duration.Duration
 import scala.concurrent.ExecutionContext
 
 object PlatformCompat {
+  def awaitResult[T](awaitable: Awaitable[T]): T = {
+    scalanative.runtime.loop()
+    Await.result(awaitable, Duration.Inf)
+  }
```
armanbilge, line 19: This is how specs2 is working as well.
https://github.com/etorreborre/specs2/blob/b4fb0a16f1a0369fe69eb994220317a5aaf2f3d9/common/native/src/main/scala/org/specs2/concurrent/package.scala#L8-L11

valencik (approved):
Published locally and tested against all our various reproducers and it worked swimmingly!

PR #703, opened 2023-09-20
Honor the NO_COLOR environment variable.

non:
As per https://no-color.org/ this change will disable ANSI colors if the NO_COLOR environment variable is set to any non-empty string.

(If the environment variable is unset or set to the empty string then the default colors will still be used.)

This builds on the work of https://github.com/scalameta/munit/pull/651 but attempts to minimize the size of the patch.


valencik (approved):
Thanks for this!
Tested with a local snapshot, things work as expected :)

PR #714, opened 2023-10-25
Use a package proxy for `scalanative.runtime.loop`

natsukagami:
...since it is becoming package-private from v0.5 onwards with https://github.com/scala-native/scala-native/pull/3423.
I'm not sure about it *staying* private though, as @wjcmazur mentioned about reverting it, so I'm keeping this PR as draft.


natsukagami:
Closing because it was made public again https://github.com/scala-native/scala-native/commit/560c4898236af130e5ad612b23d9c13b25718bc1

Issue #707, opened 2023-10-02
Drop Support for Scala 2.11


valencik:
Our Scala.js dependency bumps are failing because Scala.js dropped support for Scala 2.11 support in 1.13.0
https://github.com/scalameta/munit/pull/665#issuecomment-1656727578

We could perhaps stop publishing a Scala.js build just on 2.11, or we could drop 2.11 support altogether.

Unless there are strong reasons to continue supporting 2.11, I think we should drop it.

SethTisue:
doooooooo iiiiiiiit

tgodzik:
People can always freeze their version for 2.11.x if needed. Besides it's highly possible that munit is mostly used for newer projects anyway.

PR #723, opened 2023-12-13
Drop Support for Scala 2.11

valencik:
This PR drops support for Scala 2.11

resolves https://github.com/scalameta/munit/issues/707


tgodzik (approved):
LGTM! We need to nudge people off 2.11.x finally.

PR #724, opened 2023-12-13
Restore v0.7.x fixture ordering

valencik:
:warning: **Breaking Change** :warning: 

This PR changes the ordering in which suite-local fixture `afterAll` methods are called.
It implements the trick @olafurpg describes in https://github.com/scalameta/munit/issues/573#issuecomment-1465269356 to restore the fixture ordering of v0.7.x
This ordering was changed in the [first v1.0 milestone release](https://github.com/scalameta/munit/releases/tag/v1.0.0-M1)
Therefore it is a breaking change within the v1.0 milestones in order to avoid a breaking change between v0.7.x and v1.0 overall.

Namely, we split up the suite-local fixture implementation into two fixtures, one of the `before*` methods, and one for the `after*` methods. Then we explicitly order `munitFixtures` to have the befores up front and the afters at the end.

Of particular note, the tests/shared/src/main/scala/munit/FixtureOrderFrameworkSuite.scala file now looks the same as it did before https://github.com/scalameta/munit/pull/430/files?diff=split&w=0#diff-714f7ef13dfbbd4747770936315606f95d8aeba5f24890a45a4be216cda9d8a1

Note: The v0.7.x style ordering is not to run the `after*` methods in opposite order to how the `before*` methods were run, making a mirror image like pattern. Instead it's more like `after*` methods run in the same order as `before*` methods *except* for suite-local fixtures which are special cased to run last.

### DIFF
--- a/tests/shared/src/test/scala/munit/SuiteLocalFixtureSuite.scala
+++ b/tests/shared/src/test/scala/munit/SuiteLocalFixtureSuite.scala
@@ -12,7 +12,7 @@ class SuiteLocalFixtureSuite extends FunSuite {
       n = 1
     }
     override def afterAll(): Unit = {
-      assertEquals(n, 17)
+      assertEquals(n, 16)
       n = -11
     }
   }
@@ -34,7 +34,7 @@ class SuiteLocalFixtureSuite extends FunSuite {
   }
 
   override def afterAll(): Unit = {
-    assertEquals(counter(), 17)
+    assertEquals(counter(), -10)
   }
 
   1.to(5).foreach { i =>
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/munit/SuiteLocalFixtureSuite.scala
```diff
@@ -34,7 +34,7 @@ class SuiteLocalFixtureSuite extends FunSuite {
   }
 
   override def afterAll(): Unit = {
-    assertEquals(counter(), 17)
+    assertEquals(counter(), -10)
```
valencik, line 37: This is also reverting to it's previous value:
https://github.com/scalameta/munit/pull/430/files#diff-2ead4e24b5b9e268cfb99fcec413a0d9c860af4663e31761a3f2822dab593f39

tgodzik (approved):
Looks good to me, especially if we can keep people from being surprised by the change when going from 0.

For me a more natural way would be a complete reverse order, but that would be another breakage, so let's just go with the first approach.

olafurpg (approved):
LGTM. At this point, the 1.x series has been in milestone phase for so long that this is almost a breaking change. It's fine IMO to keep the old behavior or merge this PR as long as we communicate clearly how to get the desired ordering in the release notes.

mzuehlke (approved):

PR #727, opened 2023-12-15
chore: Don't publish the docs module

tgodzik:
I merged the update for mdoc, which uses Java 11, but that is needed to publish the website. It seems the normal release tries to publish docs and that breaks because of the mdoc JVM requirement.

This should fix the release.


tgodzik:
Let's check if that helps

tgodzik:
Ok, looks that we're good

PR #728, opened 2023-12-17
fix #712: improve assertNotEquals failure message

### LINKED ISSUE #712: `assertNotEquals` produces unhelpful message on assertion fail.
Hi,

When using `assertNotEquals` if the assertion fails, by default, the error message is quite unhelpful.
For instance with the following test (tested on munit version `1.0.0-M10`):
```scala
test("example test that fails") {
  val obtained = 42
  val expected = 42
  assertNotEquals(obtained, expected)
}
```

we obtain: `values are the same expected same: 42 was not: 42`. 

The `expected same` is a bit contradictory, same for the `was not`.  Maybe something like "values are the same,  expected two different values: 42 is equal to 42" would be enough. It's a quick fix in `Assertions.scala`, I can open a PR if needed.

ekans:
Small fix to improve failure message of `assertNotEquals` following @zaclegarssure suggestion.

### DIFF
--- a/munit/shared/src/main/scala/munit/Assertions.scala
+++ b/munit/shared/src/main/scala/munit/Assertions.scala
@@ -74,7 +74,7 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
     StackTraces.dropInside {
       if (compare.isEqual(obtained, expected)) {
         failComparison(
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM from me

PR #733, opened 2024-01-12
Upgrade Scala3 to 3.3.1

valencik:


SethTisue:
looks like overly sensitive test expectations that just need adjusting?

valencik:
It looks like the [Compare helper](https://github.com/scalameta/munit/blob/main/munit/shared/src/main/scala/munit/Compare.scala#L19-L23) isn't working on Scala 3.3.1.

For the assertion
```scala
assertEquals('a', 'a'.toInt)
```

We used to have something like:
```
error:
Can't compare these two types:
  First type:  Char
  Second type: Int
Possible ways to fix this error:
  Alternative 1: provide an implicit instance for Compare[Char, Int]
  Alternative 2: upcast either type into `Any` or a shared supertype.
```
Followed by some additional text about implicits if on Scala 3.1.

But now we have just: 
```
error: missing argument for parameter compare of method assertEquals in trait Assertions: (implicit loc: munit.Location, compare: munit.Compare[Char, Int]): Unit
```

Which seems like a regression.
I think we want to actually try and restore the old behaviour.

tgodzik:
Already updated in main

Issue #745, opened 2024-02-26
Request: Publish internal difflib as a micro-library


zainab-ali:
Munit's internal difflib is based on a [myers diff algorithm](https://github.com/scalameta/munit/blob/main/munit/shared/src/main/scala/munit/internal/difflib/MyersDiff.scala) to compare strings. 

This produces nice user-friendly diffs, even for large multi-line strings. 

It would be great to publish it as a separate library so that it can be used with other testing frameworks. For instance, it could be used alongside weaver without needing to pull in the rest of munit:

```scala
libraryDependencies += Seq(
    "org.scalameta" %% "munit-diff" % "munit-version", // only pulls in the difflib
    "com.disneystreaming" %% "weaver-cats" % "weaver-version" // weaver as the actual framework
)
```

What do you think? If this is feasible, I'd be happy to make a PR to achieve it.

tgodzik:
Thanks for reporting! We've been actually thinking about it since, we do use it in a number of places including Metals and I think scalafmt (or we would want to use it for Scala Native support).

It could even be a separate library altogether:
```
"org.scalameta" %% "scala-diff" % "version"
```
I can create a separate repo for it.

zainab-ali:
@tgodzik, that would be fantastic! Let me know once you've got the repo at hand and I'll happily help port the code over.

majk-p:
We discussed this issue offline with @tgodzik and we decided to implement `scala-diff` as a separate sbt module in `munit`. I'll pick this one

PR #756, opened 2024-03-26
Extract diff module

majk-p:
This PR aims to extract diff calculation to a separate module to resolve https://github.com/scalameta/munit/issues/745

Since diff had to share a bit of code with the rest of munit, I've also extracted `munit-core` that provides the minimal common part.


munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -288,6 +288,21 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
...
+    new ComparisonFailExceptionHandler {
+      def handle(
+          message: String,
+          obtained: String,
+          expected: String,
+          loc: Location
+      ): Nothing = {
+        assertions.failComparison(message, obtained, expected, clues)(loc)
+      }
+    }
```
majk-p, line 304: I decided to extract this from `ComparisonFailExceptionHandler` so that the `scala-diff` package doesn't need to depend on `Assertions`

majk-p:
@valencik @tgodzik this change breaks mima on purpose due to reorganisation of modules. Please advise what's your preferred strategy to deal with mima check considering 1.0.0 is underway.

build.sbt
```diff
@@ -290,6 +290,47 @@ lazy val munitScalacheck = crossProject(JSPlatform, JVMPlatform, NativePlatform)
...
+    sharedJVMSettings
+  )
+  .nativeConfigure(sharedNativeConfigure)
+  .nativeSettings(
+    sharedNativeSettings
+  )
+  .jsConfigure(sharedJSConfigure)
+  .jsSettings(sharedJSSettings)
+
+lazy val munitDiff = crossProject(JSPlatform, JVMPlatform, NativePlatform)
```
tgodzik, line 317: Ideally munitDiff would not depend on anything, it should not be about testing really.

majk-p: As far as I understand it, both diff and munit need things like `Location` along with macros that generate it. Do you prefer code duplication over shared core module?

majk-p: I think we can try getting rid of `Location`, I'll give it a try. What about `AnsiColors`? Should it be a part of diff module?

tgodzik: I think so, since it might be useful for showing the diff.

valencik:
> @valencik @tgodzik this change breaks mima on purpose due to reorganisation of modules. Please advise what's your preferred strategy to deal with mima check considering 1.0.0 is underway.

I first want to note that the ecosystem is eager for a 1.0 release.
It is not ideal to be on milestones this long.
And we have a split of projects on 0.7.x and 1.0.x milestones throughout the ecosystem.

However, I wanted to at least bump us to the latest Scala 3 LTS before cutting 1.0.
Unfortunately, we're blocked by https://github.com/scala/scala3/issues/19594 not having yet been ported back to a LTS version.

All this to say, I don't have strong objections to the plan here.

scala-diff/shared/src/main/scala/munit/diff/Diffs.scala
```diff
@@ -1,4 +1,4 @@
-package munit.internal.difflib
+package munit.diff
 
 import munit.Location
```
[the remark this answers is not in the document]
tgodzik: I think that's totally fine, asserts can live elsewhere

scala-diff/shared/src/main/scala/munit/diff/Clue.scala
```diff
@@ -0,0 +1,13 @@
+package munit.diff
+
+class Clue[+T](
```
zainab-ali, line 3: I don't think `Clue` should be a type in the diff library.  From what I understand, it's been moved here so that we can give it a custom representation in `Printers`.

That being said, I haven't thought of alternative approaches.

majk-p: Me neither, I did this to satisfy the compiler, I also don't have a better idea yet

munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -51,10 +54,10 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
       clue: => Any = "diff assertion failed"
   )(implicit loc: Location): Unit = {
     StackTraces.dropInside {
-      Diffs.assertNoDiff(
+      DiffsAssetion.assertNoDiff(
```
tgodzik, line 57:
```suggestion
      DiffsAssertion.assertNoDiff(
```

scala-diff/shared/src/main/scala/munit/diff/console/Printers.scala
```diff
@@ -0,0 +1,233 @@
...
+            else printChar(x, out)
+            out.append('\'')
+          case x: Byte   => out.append(x.toString())
+          case x: Short  => out.append(x.toString())
+          case x: Int    => out.append(x.toString())
+          case x: Long   => out.append(x.toString())
+          case x: Float  => out.append(x.toString())
+          case x: Double => out.append(x.toString())
+          case x: String => printString(x, out, printer)
+          case x: Clues =>
```
tgodzik, line 41: Could we just make Printer a class and add support for Clues in a child class?

Or maybe a custom class like:
```scala
trait CustomPrintable{
  def append(out: StringBuilder)
}
```
and make Clue implement it?

munit/shared/src/main/scala/munit/Clues.scala
```diff
@@ -2,10 +2,10 @@ package munit
...
 import munit.internal.console.Printers
 
-class Clues(val values: List[Clue[_]]) {
-  override def toString(): String = Printers.print(this)
-}
-object Clues {
-  def empty: Clues = new Clues(List())
-  def fromValue[T](value: T): Clues = new Clues(List(Clue.fromValue(value)))
-}
+// class Clues(val values: List[Clue[_]]) {
```
tgodzik, line 5: TODO remove if we decide to leave Clue in the current position

scala-diff/shared/src/main/scala/munit/diff/NextPrintable.scala
```diff
@@ -0,0 +1,8 @@
+package munit.diff
+
+/**
+ * Don't override this ever
```
tgodzik, line 4: Would be great to mention why a bit

munit/shared/src/main/scala/munit/Assertions.scala
tgodzik: Could you fix this one? It's super minor

tgodzik:
You might want to rebase the PR to see if it passes.

majk-p:
After brainstorming with @zainab-ali and inspired by [this commit](https://github.com/typelevel/weaver-test/pull/25/files/4ae049fa437fc264a25d78fc9765254722a52331#diff-fae65363220efc805c13a724f23cb62818d8347b24dfccade13a55bc91dc92ae) I did an overhaul of this change. Most of the `Printers` functionality remains in `munit` while only logic for string cleanup is extracted to `diff` (since it's the only used part). I think it makes things cleaner and we no longer have to deal with disentangling `Clues`.

scala-diff/shared/src/main/scala/munit/diff/NextPrintable.scala
majk-p: No longer relevant, shouldn't be pushed in the first place :wink:

tgodzik:
Looks like the compilation is failing, could you take a look?

majk-p:
I just run `compile` and `test` locally and everything worked well. I cannot see the CI logs in actions :disappointed:

build.sbt
```diff
@@ -214,7 +214,7 @@ lazy val munit = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(
     sharedSettings,
     Compile / unmanagedSourceDirectories ++=
-      crossBuildingDirectories("munit", "main").value,
+      crossBuildingDirectories("scala-diff", "main").value,
```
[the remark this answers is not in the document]
majk-p: That's right, missed it during cleanup

tgodzik (approved):
LGTM from me! Thanks for the great work!

I added 3 more commits:
- rename to munit-diff since we have package munit.diff everywhere so it makes much more sense
- fix mdoc docs generation
- fix mima exceptions

@zainab-ali what do you think? 

My next plan is to remove any deprecation we can find and release an RC1

zainab-ali:
Thanks @majk-p and @tgodzik, the PR itself looks great!

Unfortunately the rencent changes to munit include a Native upgrade to `0.5`. Weaver has a way to go until moving to `0.5` itself - see the [discussion for cats-effect](https://github.com/orgs/typelevel/discussions/135).

I'm brainstorming approaches we could take. Would it be viable to release this on `0.4.x` too?

tgodzik:
I can do that manually later on. Hopefully, this time it will work reasonably.

Issue #711, opened 2023-10-18
Broken compileErrors behavior on Scala 3


MateuszKubuszok:
MUnit implements `compileErrors` on Scala 3 with `inline def` delegating to `typeCheckErrors`. In the source of `typeCheckErrors` we can read:

```scala
/** Whether the code type checks in the current context? If not,
 *  returns a list of errors encountered on compilation.
 *  IMPORTANT: No stability guarantees are provided on the format of these
 *  errors. This means the format and the API may change from
 *  version to version. This API is to be used for testing purposes
 *  only.
 *
 *  An inline definition with a call to `typeCheckErrors` should be transparent.
 *
 *  @param code The code to be type checked
 *
 *  @return a list of errors encountered during parsing and typechecking.
 *
 *  The code should be a sequence of expressions or statements that may appear in a block.
 */
transparent inline def typeCheckErrors(inline code: String): List[Error] = ...
```

This lack of `transparent` is a source of unwanted behavior that I noticed in my project and, from what I heard, also other people in their projects. However, it is quite difficult to create a reproduction that is completely independent of external libraries, which is why nobody reported it yet.

I also failed to came up with a simple reproduction, but I can demonstrate the issue using existing libraries:

```scala
//> using scala 3.3.1
//> using dep io.scalaland::chimney::0.8.0
//> using dep org.scalameta::munit::1.0.0-M10
import io.scalaland.chimney.dsl.*
import munit.internal.MacroCompat

object IOnlyNeedErrors extends MacroCompat.CompileErrorMacro {

  /* Copy/Paste from munit, with transparent keyword added. */
  transparent inline def compileErrorsFixed(inline code: String): String = {
    val errors = scala.compiletime.testing.typeCheckErrors(code)
    errors
      .map { error =>
        val indent = " " * (error.column - 1)
        val trimMessage = error.message.linesIterator
          .map { line =>
            if line.matches(" +") then ""
            else line
          }
          .mkString("\n")
        val separator = if error.message.contains('\n') then "\n" else " "
        s"error:$separator$trimMessage\n${error.lineContent}\n$indent^"
      }
      .mkString("\n")
  }
}

case class Source(a: Int)
case class Target(b: String)

// If uncommented:
//Source(1).transformInto[Target]
// results in:
//Chimney can't derive transformation from Playground.Source to Playground.Target
//
//Playground.Target
//  b: java.lang.String - no accessor named b in source type Playground.Source
//
//
[...]

PR #759, opened 2024-04-15
improvement: Add transparent to compileErrors

tgodzik:
This seems to fix issues with https://github.com/scalameta/munit/issues/711


valencik (approved):
LGTM
Thanks for the reference link!

> An inline definition with a call to typeCheckErrors should be transparent.

✅

Issue #760, opened 2024-04-15
Publish for Scala Native 0.5, please


SethTisue:
this will unblock a bunch of downstream libraries from also publishing for 0.5

tgodzik:
Question, should we:
1. drop 0.4.x and release a new milestone (I intend to address the milestone elephant in the room in the near future)
2. backport 0.5.x to the latest milestone
3. Try cross publishing for 0.4 and 0.5

I can do 2. manually now to unblock downstream projects, but is it enough to do 1 later on?

valencik:
I think we're blocked by https://github.com/scala/scala3/pull/20171

tgodzik:
> > 3. Try cross publishing for 0.4 and 0.5
> 
> Definitely not this.

Yep, that's what I was hoping not to do :D

I backported munit 1.0.0-M11 for 2.13, 2.12 and 3. 

munit-scalafix requires scalafix to be released for 0.5.0

I will bump native version in the repo once that is out and next releases would be for 0.5 then.

mpilquist:
scalacheck 1.18.0 has been published for native 0.5.

tgodzik:
Thanks!

PR #761, opened 2024-04-17
chore: Update scala native to 0.5.1

tgodzik:
Fixes https://github.com/scalameta/munit/issues/760

### DIFF
--- a/tests/shared/src/test/scala/munit/AssertionsSuite.scala
+++ b/tests/shared/src/test/scala/munit/AssertionsSuite.scala
@@ -55,6 +55,7 @@ class AssertionsSuite extends BaseSuite {
   }
 
   test("false-negative") {
+    assume(BuildInfo.scalaVersion != "3.3.3")
     assertNoDiff(
       compileErrors("assertEquals(List(1), Vector(1))"),
       if (isDotty)
@@ -95,6 +96,7 @@ class AssertionsSuite extends BaseSuite {
   }
 
   test("unrelated") {
+    assume(BuildInfo.scalaVersion != "3.3.3")
     assertNoDiff(
       compileErrors("""
 class A {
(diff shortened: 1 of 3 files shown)


tgodzik:
Fixes https://github.com/scalameta/munit/issues/760

tgodzik:
This is blocked by the issue in LTS

mzuehlke:
Just thinking loud, would we loose anything if we move munit to 3.3.3 and uncomment the failing test ?
The people who are on Scala LTS and use munit would still not get the right error message.
We just have to remember to enable the tests once the fix landed in LTS

mpilquist:
At minimum, munit needs to move to 3.3.x to consume the scalacheck artifacts.

tgodzik:
Maybe yeah. I will do that in this pr

tests/shared/src/test/scala/munit/AssertionsSuite.scala
```diff
@@ -54,49 +54,50 @@ class AssertionsSuite extends BaseSuite {
...
-           |  Second type: scala.collection.immutable.Vector[Int]
-           |Possible ways to fix this error:
-           |  Alternative 1: provide an implicit instance for Compare[List[Int], scala.collection.immutable.Vector[Int]]
-           |  Alternative 2: upcast either type into `Any` or a shared supertype
-           |assertEquals(List(1), Vector(1))
-           |            ^
-           |""".stripMargin
-    )
-  }
+  if (BuildInfo.scalaVersion != "3.3.3") {
```
tgodzik, line 57: Ignored the tests specifically for 3.3.3 so this should not be ignored for the next LTS

mzuehlke: looks good 👍 

another possibility would be to have  in each of the test as first line
```Scala
assume(BuildInfo.scalaVersion != "3.3.3")
```

tgodzik: That's a nicer solution, thanks!

mzuehlke (approved):

tgodzik:
I will go ahead and merge this, then release next and hopefully last milestone

PR #768, opened 2024-04-19
Handle infinite timeouts

### LINKED ISSUE #572: setting `munitTimeout = Duration.Inf` fails
Calling `.toMillis` on a non-finite `Duration` throws an error.

https://github.com/scalameta/munit/blob/166698e290701b73e96377f57414622f2e95d530/munit/jvm/src/main/scala/munit/internal/PlatformCompat.scala#L42

So either the type should be refined to `FIniteDuration`, if that is what is expected, or this case should be specially handled.

mzuehlke:
Changing the type of `munitTimeout` from `Duration` to `FiniteDuration` would be a breaking change, so we need to accept that durations can be infinite.

Fixes #572


tgodzik (approved):
LGTM

PR #773, opened 2024-04-24
chore: Remove deprecated methods before 1.0.0

tgodzik:
I think it's high time to do it since the next option will be at 2.0.0.

After that I will do RC1, bump mima to use it and wait ~2 weeks before releasing 1.0.0?

What do people think?

### DIFF
--- a/munit/jvm/src/main/scala/munit/internal/PlatformCompat.scala
+++ b/munit/jvm/src/main/scala/munit/internal/PlatformCompat.scala
@@ -43,13 +43,7 @@ object PlatformCompat {
     task.execute(eventHandler, loggers)
     Future.successful(())
   }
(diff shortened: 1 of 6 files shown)


tgodzik:
>  But that would require us to wait for Scala 3.3.4

It's probably 4 weeks away or more if any regression sneaks in :/

mzuehlke:
> Regarding the release schedule, I would really have liked to get the implicit error messages fixed for a 1.0.0 release. But that would require us to wait for Scala 3.3.4

The PR hasn't even been merged to the LTS branch yet

mzuehlke:
> After that I will do RC1, bump mima to use it and wait ~2 weeks before releasing 1.0.0?
> 
> 
> 
> What do people think?

A 1.0 is long overdue

PR #774, opened 2024-04-26
Verify that the Scala 3 LTS bugfix backport doesn't break anything

mzuehlke:
As there isn't an artifact for Scala 3 LTS that includes https://github.com/scala/scala3/pull/20171 I gave it a try with 3.4.2-RC1 that already contains https://github.com/scala/scala3/pull/19737

This PR shows that the only thing that changes is the message returned form the compiler and not changes to munit are needed once Scala 3 LTS with the fix gets released.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ def scala213 = "2.13.13"
 
 def scala212 = "2.12.19"
 
-def scala3 = "3.3.3"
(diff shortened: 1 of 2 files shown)


tgodzik:
The PR should be in the next LTS, I confirmed it with the team, they just have a lot on their plate currently.

PR #776, opened 2024-04-29
chore: Reenable mima to make sure we don't break compat

tgodzik:


valencik:
I'm pretty unfamiliar with how this is normally done.
It looks like we are wiping out all past MIMA exclusions and starting fresh with the 1.0 release?
Is that the common practice?

tgodzik:
> I'm pretty unfamiliar with how this is normally done. It looks like we are wiping out all past MIMA exclusions and starting fresh with the 1.0 release? Is that the common practice?

I think so, since we no longer need to compare to 0.x versions, just 1.x. The old exclusions are irrelevant now. We should not have any exclusions aside from added methods/classes on the 1.x series.

mzuehlke (approved):
> I think so, since we no longer need to compare to 0.x versions, just 1.x. The old exclusions are irrelevant now. We should not have any exclusions aside from added methods/classes on the 1.x series.

Sounds legit.

PR #766, opened 2024-04-18
Align output for ignored and skipped tests between all 3 platforms

mzuehlke:
This aligns the out of ignore and skipped `assume(false)`test on all 3 platforms.

```
munit.SkippedFrameworkSuite:
  + pass <elapsed time>
==> i ignore ignored
  + assume(true) <elapsed time>
==> s assume(false) skipped
```

I added tests for the behavior of the verbose and the default mode on the JVM and on JS + Native (there are still small differences).

The first commit in this PR just introduces the test with the prior behavior.
The 2nd commit introduces the change.

Fixes #549

PR #779, opened 2024-05-09
Don't print test suite name if no test is being run

alexarchambault:
These changes make munit print only the test suite names of suites for which it actually ran tests, rather than printing all suite names (even those for which all tests are being filtered out), as it's been doing up to now.

For example, instead of printing things like
```text
$ ./mill -i -w scala.integration.test "almond.integration.KernelTestsSimple3.completion"
…
almond.integration.KernelTestsSimple3:
  + completion 17.073s
almond.integration.KernelTestsTwoStepStartup3:
almond.integration.KernelTestsSimple213:
almond.integration.KernelTestsTwoStepStartup212:
almond.integration.KernelTestsTwoStepStartup213:
almond.integration.KernelTestsSimple212:
```

it now prints
```text
$ ./mill -i -w scala.integration.test "almond.integration.KernelTestsSimple3.completion"
…
almond.integration.KernelTestsSimple3:
  + completion 15.468s
```

I find the latter to be more readable.

### DIFF
--- a/munit/shared/src/main/scala/munit/MUnitRunner.scala
+++ b/munit/shared/src/main/scala/munit/MUnitRunner.scala
@@ -122,15 +122,20 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
   }
   def runAsync(notifier: RunNotifier): Future[Unit] = {
     val description = getDescription()
-    notifier.fireTestSuiteStarted(description)
-    runAll(notifier)
-      .transformCompat[Unit](result => {
-        result.failed.foreach(ex =>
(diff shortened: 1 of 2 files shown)


alexarchambault:
Not sure how to test that, or if it needs to be tested. (Going to see if it break things on the CI here…)

Manual tests show that it works as expected, at least.

tgodzik (approved):
This looks good to me, though I wonder if there was no initial purpose for printing thise? @valencik  @mzuehlke any opinions here?

mzuehlke (approved):
I made the output of individual tests that are ignored or skipped more verbose in #766.
But for complete suites I find your change good.👍🏻 
What are scalatest or junit doing in these cases?

PR #788, opened 2024-05-27
Add `contextSize` test parameter to enlarge the scope of a diff

TonioGela:
As per description, this PR introduces a `contextSize` overridable suite parameter that control the number of lines around the reported diff in case of a mismatch ( ala `grep -A <n> -B <n>`)

Ofc the normal behaviour was kept identical to the actual, and there's no change in the user API (at least for the methods used via `Assertions`)

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -196,6 +196,83 @@ List(
 ...
 ```
 
+## Enlarging the diff context
+
+When a test fails Munit shows a rich difference, to easily spot the difference between the test result and the obtained value.
+Understanding where the reported value actually is can sometimes be difficult, specifically with repeated values, as by default Munit shows the diff surrounding it just with a single line of context.
+
+The context shown in the diff can be enlarged in a specific testing suite by overriding `contextSize`.
+
+```scala mdoc
+import munit.FunSuite
+
+class CustomContextSizeTest extends FunSuite {
+  override def contextSize: Int = 10
+
+  test("contextSize") {
+    val a = List("a", "a", "a", "a", "a", "a", "a", "a", "a")
+    val b = List("a", "a", "a", "a", "b", "a", "a", "a", "a")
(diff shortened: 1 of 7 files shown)


TonioGela:
Hmm, given that a lot of the stuff in `Diff` is effectively exposed the mima check screams with 11 compatibility errors. I'll try to figure out a better encoding for the content of this PR, so don't take it too seriously for the moment 😇

tgodzik:
Shouldn't this be a parameter to the runner https://github.com/scalameta/munit/blob/8b63f0e7f8574cd7910816d171d98bbf2507d672/munit/js-native/src/main/scala/munit/internal/junitinterface/JUnitFramework.scala#L110 ?

TonioGela:
> Shouldn't this be a parameter to the runner
> 
> https://github.com/scalameta/munit/blob/8b63f0e7f8574cd7910816d171d98bbf2507d672/munit/js-native/src/main/scala/munit/internal/junitinterface/JUnitFramework.scala#L110
> ?

Interesting question. Going further: should it just be a parameter or also a parameter?

tgodzik:
I think just a parameter since it doesn't really change the test logic, but how the test is being displayed. Potentially we could have runner defaults for a specific suite as a separate thing.

tgodzik:
btw. MiMa is not a huge issue here, we would need to do 1.1.x next so that it's automatically picked up as possible incompatibility

TonioGela:
> I think just a parameter since it doesn't really change the test logic, but how the test is being displayed. Potentially we could have runner defaults for a specific suite as a separate thing.

Okay, this seems reasonable. I'll try to rework it. I'll also try to add documentation about that flag maybe here: https://scalameta.org/munit/docs/troubleshooting.html

kitbellew:
@TonioGela please take a look at #883 and see if there's anything else left here.

Issue #382, opened 2021-06-22
munit.Location incompatible with Gradle build caching and Distributed Testing


DanielThomas:
The compiler macro embeds the fully qualified path to the source file for location at test runtime, however this causes the output to not be portable between machines, making it incompatible with Gradle build caching (or anything that wants to ship test artifacts outside the immediate build environment for that matter).

This also means that tests can't be distributed using the distributed test mechanism, as all resources need to be available via the classloader:

https://docs.gradle.com/enterprise/test-distribution-gradle-plugin/#tests_reading_files_from_src_test_resources

Attempting to locate source files as relative classloader resources before falling back to the file system would provide the flexibility necessary to implement support for both of these features downstream, and allow tests to be packaged with their sources in a standalone jar for running outside of the build tool.

eed3si9n:
Here's my PR for this - https://github.com/scalameta/munit/pull/823

PR #823, opened 2024-09-10
Fix Location macro to be hermetic

eed3si9n:
Fixes https://github.com/scalameta/munit/issues/382

## Problem
The `Location` macro currently captures the absolute path of the source file, which is not hermetic, and won't cache correctly on remote cached build tools such as Bazel, Gradle, and (in development) sbt 2.x.

```diff
$ jardiff target/out/jvm/scala-3.4.2/inctest/test-backend/ ../inctest2/target/out/jvm/scala-3.4.2/inctest/test-backend/
diff --git a/example/CowTest.class.asm b/example/CowTest.class.asm
index 1b3911e..91fd219 100644
--- a/example/CowTest.class.asm
+++ b/example/CowTest.class.asm
@@ -56,7 +56,7 @@
     ]
     NEW munit/Location
     DUP
-    LDC "/Users/xxx/work/inctest/src/test/scala/example/CowTest.scala"
+    LDC "/Users/xxx/work/inctest2/src/test/scala/example/CowTest.scala"
     BIPUSH 6
     INVOKESPECIAL munit/Location.<init> (Ljava/lang/String;I)V
     INVOKEVIRTUAL example/CowTest.assert (Lscala/Function0;Lscala/Function0;Lmunit/Location;)V
```

## Solution
This captures the relative path from the working directory instead.

### DIFF
--- a/munit/shared/src/main/scala/munit/internal/console/Lines.scala
+++ b/munit/shared/src/main/scala/munit/internal/console/Lines.scala
@@ -23,7 +23,10 @@ class Lines extends Serializable {
       val somePath1 =
         if (somePath.endsWith(sep)) somePath.dropRight(sep.length)
         else somePath
-      somePath1.split(sep).dropRight(1).mkString(sep)
+      val sep1 =
+        if (sep == "\\") "\\\\"
+        else sep
+      somePath1.split(sep1).dropRight(1).mkString(sep)
     }
     if (Files.exists(p)) p
     else if (max < 1) sys.error(s"$path was not found")
--- a/munit/shared/src/main/scala/munit/internal/console/Lines.scala
+++ b/munit/shared/src/main/scala/munit/internal/console/Lines.scala
@@ -16,9 +16,18 @@ class Lines extends Serializable {
   }
   def formatPath(location: Location): String =
     location.path
+
+  def findPath(cwd: String, path: String, max: Int): Path = {
+    val p = Paths.get(cwd).resolve(path)
+    if (Files.exists(p)) p
+    else if (max < 1) sys.error(s"$path was not found")
+    else if (cwd.contains("\\")) findPath(cwd.split("\\").dropRight(1).mkString("\\"), path, max - 1)
+    else findPath(cwd.split("/").dropRight(1).mkString("/"), path, max - 1)
+  }
+
   def formatLine(location: Location, message: String, clues: Clues): String = {
     try {
-      val path = Paths.get(location.path)
+      val path = findPath(Path.workingDirectory.toString, location.path, 3)
       val lines = filecache.getOrElseUpdate(
         path, {
           Files.readAllLines(path).asScala.toArray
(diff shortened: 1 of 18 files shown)


munit/shared/src/main/scala/munit/internal/console/Lines.scala
```diff
@@ -16,9 +16,18 @@ class Lines extends Serializable {
   }
   def formatPath(location: Location): String =
     location.path
+
+  def findPath(cwd: String, path: String, max: Int): Path = {
+    val p = Paths.get(cwd).resolve(path)
+    if (Files.exists(p)) p
+    else if (max < 1) sys.error(s"$path was not found")
+    else if (cwd.contains("\\")) findPath(cwd.split("\\").dropRight(1).mkString("\\"), path, max - 1)
```
tgodzik, line 24: Looks like some of the tests are failing on Windows, would you be able to take a look?

tgodzik (changes requested):

munit/shared/src/main/scala/munit/internal/console/Lines.scala
mzuehlke: Scalafmt and Scalafix are having an issue, too

eed3si9n: I guess it's failing because the expected value contains / and on Windows it's returning \.

munit/shared/src/main/scala/munit/internal/console/Lines.scala
```diff
@@ -19,11 +19,17 @@ class Lines extends Serializable {
...
+        if (somePath.endsWith(sep)) somePath.dropRight(sep.length)
+        else somePath
+      somePath1.split(sep).dropRight(1).mkString(sep)
+    }
     if (Files.exists(p)) p
     else if (max < 1) sys.error(s"$path was not found")
     else if (cwd.contains("\\"))
-      findPath(cwd.split("\\").dropRight(1).mkString("\\"), path, max - 1)
-    else findPath(cwd.split("/").dropRight(1).mkString("/"), path, max - 1)
+      findPath(getParentPath(cwd, "\\"), path, max - 1)
```
mzuehlke, line 31: What do you think about using https://docs.oracle.com/javase/8/docs/api/java/io/File.html#separator instead of looking at the current path ? Or would this only work on the JVM ?

eed3si9n: Yea, for native and JS it might be simpler to just look for `\`.

tgodzik (approved):
LGTM!

tgodzik:
Let me know if you need a release right away.

eed3si9n:
It would be great if could get a release this week or next. Thanks!

tgodzik:
It's running! https://github.com/scalameta/munit/actions/runs/10847736559

PR #837, opened 2024-10-03
Remove Stability section because v1.x was released

seblm:
It is a simple change that remove Stability section of the documentation because project is now published with v1.x tags.


tgodzik (approved):
Thanks!

Issue #782, opened 2024-05-11
Full stack trace for assertEquals


timothyklim:
For `assert` I get full stack trace like
```
==> X xxx.XxxSpec.xxx  0.003s munit.FailException: assertion failed
    at munit.FunSuite.assert(FunSuite.scala:11)
    at xxx.XxxSpec.$init$$$anonfun$2$$anonfun$1(XxxSpec.scala:72)
    at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619)
    at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617)
    at scala.collection.parallel.immutable.ParHashMap$ParHashMapIterator.foreach(ParHashMap.scala:78)
    at scala.collection.parallel.ParIterableLike$Foreach.leaf(ParIterableLike.scala:938)
    at scala.collection.parallel.Task.tryLeaf$$anonfun$1(Tasks.scala:52)
    at scala.util.control.Breaks$$anon$1.catchBreak(Breaks.scala:97)
    at scala.collection.parallel.Task.tryLeaf(Tasks.scala:56)
    at scala.collection.parallel.Task.tryLeaf$(Tasks.scala:21)
    at scala.collection.parallel.ParIterableLike$Foreach.tryLeaf(ParIterableLike.scala:935)
    at scala.collection.parallel.AdaptiveWorkStealingTasks$AWSTWrappedTask.internal(Tasks.scala:159)
    at scala.collection.parallel.AdaptiveWorkStealingTasks$AWSTWrappedTask.internal$(Tasks.scala:142)
    at scala.collection.parallel.AdaptiveWorkStealingForkJoinTasks$AWSFJTWrappedTask.internal(Tasks.scala:304)
    at scala.collection.parallel.AdaptiveWorkStealingTasks$AWSTWrappedTask.compute(Tasks.scala:149)
    at scala.collection.parallel.AdaptiveWorkStealingTasks$AWSTWrappedTask.compute$(Tasks.scala:142)
    at scala.collection.parallel.AdaptiveWorkStealingForkJoinTasks$AWSFJTWrappedTask.compute(Tasks.scala:304)
    at java.util.concurrent.RecursiveAction.exec(RecursiveAction.java:194)
    at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
    at java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1491)
[...]

PR #856, opened 2024-11-28
bugfix: Print stack trace for assertEquals

tgodzik:
Not sure why not show the stack trace and it's rather unexpected

Fixes https://github.com/scalameta/munit/issues/782

Opinions @valencik @mzuehlke ?


mzuehlke (approved):
The 2 test cases you had to adapt didn't even made sense before.
Both didn't contain a stacktrace.

So I think this a good change

AugustNagro:
Awesome, thanks!

Issue #868, opened 2025-01-15
Build of IgnoreSuite uses the latest class file version supported by the JDK


cushon:
When building with newer JDK versions, the class file outputs for Scala files in this project target JDK 8 bytecode, but Java files like `./munit/jvm/src/main/java/munit/IgnoreSuite.java` are being built with the default javac flags and use the latest target version supported by the JVM.

Should the build be configured to use a consistent target version for the Java files?

Demo:

```
$ java -version
openjdk version "23.0.1" 2024-12-05
$ sbt compile
$ javap -v -p ./munit/jvm/target/scala-2.13/classes/munit/IgnoreSuite.class
...
  major version: 67
```

PR #869, opened 2025-01-15
Consider a -source and -target version for javac

cushon:
This avoids building e.g. IgnoreSuite targeting the latest class file version supported by the JDK used at build time.

https://github.com/scalameta/munit/issues/868


tgodzik (approved):
Makes sense, it's safer to specify those than relying on having to run on the specific Java version

cushon:
> Could you rebase on main? I fixed the CI as it was missing sbt due to new github images being rolled out

Done, thanks!

PR #875, opened 2025-01-18
Diff: swap the order of expected-vs-obtained

kitbellew:
It was mapping obtained to original, and expected to revised.


tgodzik (approved):
I think it makes more sense, yeah. Though wonder if it could confuse people that expect the old order?

kitbellew:
> I think it makes more sense, yeah. Though wonder if it could confuse people that expect the old order?

i would emphasize this in the release notes. this particular approach took me a while to get used to, but the last straw was a recent conversation with @bjaglin where i completely misunderstood his bug report where he shared a simple diff between what was checked in and the actual output.

SethTisue:
This has always subconsciously bugged me, but I never thought to suggest it be changed. Thank you!

PR #883, opened 2025-01-20
Diff: add new DiffOptions parameter

kitbellew:
Sets diff context size and optionally outputs line offsets. This is part 2, following #881. Builds upon #788.

### DIFF
--- a/munit-diff/shared/src/main/scala/munit/diff/DiffOptions.scala
+++ b/munit-diff/shared/src/main/scala/munit/diff/DiffOptions.scala
@@ -1,0 +1,31 @@
+package munit.diff
+
+class DiffOptions private (
+    val contextSize: Int,
+    val showLines: Boolean,
+    val obtainedAsStripMargin: Boolean,
+) {
+  private def privateCopy(
+      contextSize: Int = this.contextSize,
+      showLines: Boolean = this.showLines,
+      obtainedAsStripMargin: Boolean = this.obtainedAsStripMargin,
+  ): DiffOptions = new DiffOptions(
+    contextSize = contextSize,
+    showLines = showLines,
+    obtainedAsStripMargin = obtainedAsStripMargin,
+  )
+
+  def withContextSize(value: Int): DiffOptions = privateCopy(contextSize = value)
+  def withShowLines(value: Boolean): DiffOptions = privateCopy(showLines = value)
+  def withObtainedAsStripMargin(value: Boolean): DiffOptions =
+    privateCopy(obtainedAsStripMargin = value)
+}
+
+object DiffOptions
+    extends DiffOptions(
+      contextSize = 1,
+      showLines = false,
+      obtainedAsStripMargin = false,
+    ) {
+  implicit val default: DiffOptions = this
+}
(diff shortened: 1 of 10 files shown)


munit-diff/shared/src/main/scala/munit/diff/Diffs.scala
```diff
@@ -1,5 +1,6 @@
 package munit.diff
 
+@deprecated("Please use Diff instead", "1.0.4")
```
tgodzik, line 3:
```suggestion
@deprecated("Please use Diff instead", "1.1.0")
```
I would say if we are adding anything to API we need a new minor

tests/shared/src/main/scala/munit/Issue179FrameworkSuite.scala
```diff
@@ -13,8 +13,8 @@ object Issue179FrameworkSuite
          |5:}
          |diff assertion failed
          |=> Obtained
-         |    '''|
```
tgodzik, line 16: Why the change, I think should stay, since it was there to easily copy or see what is going on

munit-diff/shared/src/main/scala/munit/diff/DiffOptions.scala
```diff
@@ -0,0 +1,31 @@
...
+  def withShowLines(value: Boolean): DiffOptions = privateCopy(showLines = value)
+  def withObtainedAsStripMargin(value: Boolean): DiffOptions =
+    privateCopy(obtainedAsStripMargin = value)
+}
+
+object DiffOptions
+    extends DiffOptions(
+      contextSize = 1,
+      showLines = false,
+      obtainedAsStripMargin = false,
```
tgodzik, line 28:
```suggestion
      obtainedAsStripMargin = true,
```
to keep consistent with what was before, no?

kitbellew: turns out the "before" was a bit inconsistent. there were several cases when it was true but more cases when it was false. at least the number of failing tests with true was a lot larger than with false.

munit-diff/shared/src/main/scala/munit/diff/Diffs.scala
kitbellew: happy to change, but they way i have been interpreting this (and using in scalameta) is using the last version where this functionality was NOT deprecated (since that's basically the current version, which is known), instead of the first version where it will be deprecated (which is the next version, and yet to be decided).

so, 1.0.4 is the last released one. please confirm you'd like me to change to what we think the next one will be.

tests/shared/src/main/scala/munit/Issue179FrameworkSuite.scala
kitbellew: i can provide a custom implicit instance of `DiffOptions` within this test which would preserve this. see comment above, about inconsistency of `printObtainedAsStripMargin`.

munit-diff/shared/src/main/scala/munit/diff/Diffs.scala
tgodzik: Ach, right. I thought we released 1.0.3 last. my bad. This is fine.

munit-diff/shared/src/main/scala/munit/diff/DiffOptions.scala
tgodzik: ach, ok. makes sense to have a smaller impact.

tgodzik (approved):
LGTM then, thanks for explaning!

PR #885, opened 2025-01-21
DiffOptions: allow controlling use of ANSI markers

kitbellew:
Also, exclude `munit.internal.` from MIMA.

### DIFF
--- a/munit/shared/src/main/scala/munit/Assertions.scala
+++ b/munit/shared/src/main/scala/munit/Assertions.scala
@@ -3,7 +3,6 @@ package munit
 import munit.diff.DiffOptions
 import munit.diff.EmptyPrinter
 import munit.diff.Printer
-import munit.diff.console.AnsiColors
 import munit.internal.MacroCompat
 import munit.internal.console.Lines
 import munit.internal.console.Printers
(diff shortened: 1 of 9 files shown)


munit/shared/src/main/scala/munit/Assertions.scala
```diff
@@ -228,27 +226,61 @@ trait Assertions extends MacroCompat.CompileErrorMacro {
...
+      diffOptions: DiffOptions,
+  ): Nothing = throw new FailException(
+    munitLines.formatLine(loc, message, ansi = useAnsiColors),
+    cause,
+    isStackTracesEnabled = true,
+    location = loc,
+  )
+
+  // for MIMA compatibility
+  @deprecated("Use version with implicit DiffOptions", "1.0.4")
```
tgodzik, line 254: I don't think it will work, though if someone is overriding the method it will show an error.

kitbellew: this provides binary compatibility (the parameters, same order) except I'm not sure if private is sufficient, too; and the new method is source-compatible.

tgodzik: Though if we override it wil fail the compilation or am I missing something:

```
trait Location
trait A{
  def locs(a: String, b: Int, l: Location): Nothing = ???
}

class B extends A{
  override def locs(a : String, b: Int)(implicit l : Location) = ???
}
```

This will fail, so I assume your change should also fail

tgodzik: I would think so, which is why I am bit confused. It might be a bug in Mima or alternatively the code produced is binary compatible, just not source compatible.

kitbellew: yes, mima checks for binary. well, if there's minor source incompatibility, increasing the minor version, as you suggested, should be enough. wdyt?

tgodzik (approved):

kitbellew:
@tgodzik thank you. i have no more changes to propose, should i release 1.1.0?

tgodzik:
> @tgodzik thank you. i have no more changes to propose, should i release 1.1.0?

Nothing from me.

PR #919, opened 2025-04-30
Fix munit-scalacheck version

lenguyenthanh:
The displayed version in the website is `1.0.4` (a.k.a `STABLE_VERSION`) which doesn't even exist


tgodzik:
Thanks!

PR #926, opened 2025-05-20
MunitRunner: cosmetic, refactor before-{each,all}

kitbellew:


munit/shared/src/main/scala/munit/MUnitRunner.scala
```diff
@@ -180,12 +179,15 @@ class MUnitRunner(val cls: Class[_ <: Suite], newInstance: () => Suite)
       )
     )
     result.map { results =>
-      val loadedFixtures = results.collect { case Success((fixture, true)) =>
-        fixture
+      val loadedFixtures = List.newBuilder[AnyFixture[_]]
```
mzuehlke, line 182: sadly `partitionMap` is only available from 2.13 on

PR #948, opened 2025-05-30
chore: Switch to maven central to fix release

tgodzik:


kitbellew (approved):

kitbellew:
@tgodzik looks like with the new version of sbt-ci-release we are publishing only 2.13 versions, at least for the snapshot release. either we need to change something in the `build.sbt` file, or perhaps changes to support snapshots in central repo were incomplete.

@eed3si9n any advice?

https://github.com/scalameta/munit/actions/workflows/release.yml: last two builds are using 1.11.0 (publishing just 2.13), and the two just before 1.9.3 (contain "Setting Scala version to..." messages and multiple cross versions).

PR #967, opened 2025-07-22
Do not swallow exceptions in afterEach. Fixes #966

### LINKED ISSUE #966: Do not swallow exceptions in afterEach
It appears to me that munit currently swallows all exceptions that occur in `afterEach`. Is this the desired behaviour? It was very surprising to me and imo is rarely what you want. The clean-up code is there for a reason and often affects the outcome of following tests. 

In a scala-lang forum thread, one of the main contributors of this project [suggested to use `fail` or `failSuite` to abort](https://contributors.scala-lang.org/t/announcing-munit-a-new-scala-testing-library/4007/9#:~:text=That%E2%80%99s%20an%20interesting%20idea%2C%20MUnit%20has%20a%20fail()%20helper%20to%20fail%20a%20single%20test%20case%20and%20you%20can%20call%20that%20in%20the%20afterEach%20method.) when an exception is thrown, but this is also not possible because `failSuite`, too, merely throws a special exception that will be swallowed.

At the very least, the exception should be logged and ideally the affected test or the entire suite should be failed with status "assumption failed".

### Implementation hints:

This seems to be the line where the exception is swallowed because the result `Try` is ignored in `transformWith`:
https://github.com/scalameta/munit/blob/50aa2fff7880292bbfa7d6a0476270c8fe7ff28b/munit/shared/src/main/scala/munit/MUnitRunner.scala#L276

The test can be aborted with `notifier.fireTestAssumptionFailed`. If the test has a failed outcome, then any exception in the `afterEach` method can be added to the failure as a suppressed exception.

tschuchortdev:
This should fix the issue in #966. I couldn't figure out how to write a test for it. Someone should probably do that before merging to make sure that it works as expected.

### DIFF
--- a/munit/shared/src/main/scala/munit/MUnitRunner.scala
+++ b/munit/shared/src/main/scala/munit/MUnitRunner.scala
@@ -269,23 +269,59 @@ class MUnitRunner(val cls: Class[_ <: Suite], suite: Suite)
       notifier: RunNotifier,
       description: Description,
       test: Test,
-  ): Future[Unit] = runBeforeEach(test).flatMap[Any] { beforeEach =>
-    def afterEach() = runAfterEach(test, beforeEach.loadedFixtures)
-    beforeEach.errors match {
-      case Nil => StackTraces.dropOutside(test.body())
-          .transformWith(result => afterEach().transform(_ => result))
-      case error :: errors =>
-        afterEach()
-        errors.foreach(err => error.addSuppressed(err))
-        Future.failed(error)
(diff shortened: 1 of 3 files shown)


tgodzik:
Thanks for the contribution! Did you try writing something similar to Issue285FrameworkSuite? That's where I would start.

tschuchortdev:
I have added a test based on `SwallowedExceptionSuite`. Note that I also made some changes to other files to ensure that the suite is aborted as expected, so those files will have to be reviewed again.

PR #1091, opened 2026-05-21
fix: repair broken images on landing page

iamezcua-dev:
## Problem
Several images on https://scalameta.org/munit/ fail to load. The landing page references GitHub blob URLs that return HTML, not raw image bytes.

## Fix
Updated image URLs in `website/pages/en/index.js` to use `raw.githubusercontent.com` so the browser receives the actual image content.

## Verification
- Ran `sbt docs/mdoc` + `npm start` in `website/`.
- Confirmed every image in the "Source locations for errors" section (and others previously broken) now renders locally at http://localhost:3000/munit/.


tgodzik (approved):
Thanks!

PR #1094, opened 2026-06-02
Reintroduce `@EnableReflectiveInstantiation` to Scala Native test suites via `org.portable-scala:portable-scala-reflect`

### LINKED ISSUE #1093: Regression: Munit 1.3.1 seems to silently skip tests on Scala Native
## Versions
Munit 1.3.1 (works with 1.3.0)
Scala Native 0.5.12

## Reproduction steps

With Scala CLI v1.14.0:
```scala
//> using dep org.scalameta::munit::1.3.1
//> using platform native
  test("foo") {
    assert(2 + 2 == 4)
    println("Hello from native tests")
  }
}
```

Tests are silently not run, the build succeeds (so CIs likely won't notice).
Reverting to MUnit 1.3.0 fixes it.

The same happens with SBT 1.12.11.
Only Scala Native builds are affected.

## Expected result
Tests should be detected and run, like with 1.3.0 and before

## Additional notes
- the likely culprit is https://github.com/scalameta/munit/pull/1092
- did some investigation (initially suspected it's something that Scala CLI may be doing wrong) and my current guess is that`@EnableReflectiveInstantiation` is doing the magic that's missing here

cc @zainab-ali @WojciechMazur

Gedochao:
Fixes #1093

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,6 +13,7 @@ def scala3 = "3.3.7"
 def scala3next = "3.8.2"
 
 def junitVersion = "4.13.2"
+def portableScalaReflectVersion = "1.1.3"
 def gcp = "com.google.cloud" % "google-cloud-storage" % "2.64.0"
 
 inThisBuild {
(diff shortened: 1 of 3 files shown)


WojciechMazur (approved):
LGTM

That's in fact a correct solution to the problem. The SN `test-interface` shipped deprecated shim for compatibility with Scala.js (pre 0.4.0) 

https://github.com/scala-native/scala-native/blob/0dc2c0021ff0e355332913d8e1fee4359a3d1199/test-interface/src/main/scala/scala/scalajs/reflect/annotation/package.scala#L1-L11 

It was deprecated, and the correct usage was to use portable-scala-reflect which would work as a facade for native Scala.js and Scala Native annotations. 
To be honest I'm suprised it was not found earlier, there was no deprecation warnings, and what's the most important that it continued to compile after switching dependency

PR #1092, opened 2026-05-28
Use Scala Native `test-interface-sbt-defs` instead of `test-interface`

zainab-ali:
See https://github.com/scala-native/scala-native/issues/4844

Testing libraries and frameworks should depend on Scala Native's `test-interface-sbt-defs`, not `test-interface`.
 - `test-interface-sbt-defs` defines a common interface for unit tests. It follows `early-semver` versioning.
 - `test-interface` is a server implementation that must match the version of the Scala Native plugin. It follows `strict` versioning.

This was picked up by SBT [1.12.10](https://eed3si9n.com/sbt-1.12.10), which added eviction checks for test dependencies.

## How to reproduce

Create an SBT project with the following:

```scala
// project/build.properties
sbt.version = 1.12.11
```

```scala
// project/plugins.sbt
addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.5.11")
```

```scala
// build.sbt
scalaVersion := "3.3.7"
enablePlugins(ScalaNativePlugin)

ThisBuild / evictionWarningOptions := (ThisBuild / evictionWarningOptions).value
  .withConfigurations(List(Compile, Test))

libraryDependencies += "org.scalameta" %%% "munit" % "1.0.4" % Test
```

The `sbt update` command gives:

```
[error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error]
[error] 	* org.scala-native:test-interface_native0.5_3:0.5.11 (strict) is selected over 0.5.6 for test
[error] 	    +- default:scala-native-seed-project_native0.5_3:0.1.0-SNAPSHOT (depends on 0.5.11)
[error] 	    +- org.scalameta:munit_native0.5_3:1.0.4              (depends on 0.5.6)

```

PR #1104, opened 2026-06-08
Fix test discovery for Scala 2 with Scala Native

### LINKED ISSUE #1093: Regression: Munit 1.3.1 seems to silently skip tests on Scala Native

Gedochao:
Fixes https://github.com/scalameta/munit/issues/1093
Follow-up to https://github.com/scalameta/munit/pull/1094
Fixes another failure introduced by https://github.com/scalameta/munit/pull/1092

the TL;DR of it is that https://github.com/scalameta/munit/issues/1093#issuecomment-4648687124 still happens for Scala 2.13, specifically.
The `org.portablescala.reflect.annotation.EnableReflectiveInstantiation` is not inherited there properly, while `scala.scalanative.reflect.annotation.EnableReflectiveInstantiation` seems to work fine.

Changes in this PR
- some missing `.gitignore` entries I ran into
- the actual fix (swapping `org.portablescala.reflect.annotation.EnableReflectiveInstantiation` to `scala.scalanative.reflect.annotation.EnableReflectiveInstantiation` for Scala 2
  - related unit tests
- full test discovery tests for the whole Scala version (2.12, 2.13, 3.3, 3.next) x platform (js, native) matrix ran on the CI, to prevent this problem from coming back without us knowing

cc @tgodzik @WojciechMazur


tgodzik (approved):
LGTM

dubinsky:
> * full test discovery tests for the whole Scala version (2.12, 2.13, 3.3, 3.next) x platform (js, native) matrix ran on the CI, to prevent this problem from coming back without us knowing

Thank you!!!

PR #1134, opened 2026-07-21
Improve contributing experience

OlegYch:
- add and document a command to prepare changes for a PR
- move scalafmt check to an earlier CI step for faster feedback in common scenarious


tgodzik (approved):
good idea, thanks!

PR #1146, opened 2026-08-02
CI: pack the checks into balanced workers

kitbellew:
Slots are the bottleneck during a dependency-update storm: runs spend 83% of their wall clock queued. Two pre-merge workers instead of 19, with Windows and JDK 25 deferred to push, halve the median wall clock.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,117 +9,122 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+# pr-* jobs have no event guard, so post-* only ever ADD to them; the two
+# sets cannot drift. Workers are balanced on purpose — slots, not minutes,
+# are scarce during an update storm — so keep them even when adding a step.
+
+env:
+  java: '17' # the floor, and what we release from; post-jdk25 overrides it
+
+# A job's `name` is the required-check context; renaming one needs an admin
+# change to branch protection.
 jobs:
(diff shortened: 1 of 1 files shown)


kitbellew:
@tgodzik ptal. also, this change requires removing required checks in branch protection settings, i don't have privileges to do so.

i have been changing these jobs around, since running 19 separate workers is expensive in terms of time and worker availability, especially when dependabot decides to upgrade many repos at the same time.

.github/workflows/ci.yml
```diff
@@ -9,117 +9,115 @@ concurrency:
...
-          - ubuntu-latest
-          - windows-latest
-        sbt:
-          - 'testsJVM'
-          - 'testsJS'
-          - 'testsNative'
-          - 'plugin'
-    runs-on: ${{ matrix.os }}
+  # ===== PR gate: 2 workers, run on PR *and* push =====
+  pr-jvm-native-lint:
```
tgodzik, line 23: We could probably add a name for a better explanation of what is running here

tgodzik (changes requested):
One minor comment, otherwise looks good. I think it's worth adding a name for those workflows to be clearer what they test

.github/workflows/ci.yml
kitbellew: done

kitbellew: also, i would frankly remove check requirements, we always run everything anyway :)

Issue #1161, opened 2026-08-10
Support for `eventually`


bdmendes:
specs2 has support for retrying an assertion until it succeeds or some exit condition happens (timeout/max retries): https://etorreborre.github.io/specs2/guide/SPECS2-5.9.1/org.specs2.guide.Matchers.html. I find that it is rather convenient and might fit in munit.

Is this something you'd like to include? I am thinking

```scala
def eventually(body: => Unit): Future[Unit]
```

But I am not sure if 1) it would be very clear that this is an "awaitable" and must be composed at the return value of an example or 2) it makes sense to provide this at the base library level considering CE and friends would prefer IO and friends.
An alternative is to make it return Unit and somehow block, but that might not be in the spirit of the said effect frameworks.
What's your opinion?

tgodzik:
MAybe `assertEventually` ? That fits more with the current assertions. The only issue I see is that we don't have any assertions that return Future, which might be hard for users to notice. Maybe we could do AsyncAssertions? Or just indeed block, since we are using future anyway, so that is already a problem for some frameworks, no?

bdmendes:
`assertEventually` looks okay to me, but less composable than `eventually(...)`, since we would have to replicate it for each assertion: `assertMatchesEventually`, `assertEqualsEventually`... Or lose some expression power.

> Or just indeed block, since we are using future anyway

I think so? In test frameworks they provide a `ValueTransform` for the effect type, so if they do
```scala
for
  a <- someEffect
  b <- someOtherEffect
yield eventually(assertEquals(a,b))
```
The final condition already runs in the desired thread/fiber/... as a map continuation. But then again, it does not make much sense, since `a` and `b` could never change here. (also a concern for plain Future)
To align with those I think
```scala
eventually:
  for
    a <- someEffect
    b <- someOtherEffect
  yield assertEquals(a,b)
```
But then eventually should return the effect type for it to run through the transform.
How about a new `EventuallyTransform` that knows how to retry the effect type? (implementors could do something like [this](https://github.com/cb372/cats-retry), only retrying after a proper munit failure type, etc)

tgodzik:
Och, EventuallyTransform might be a better idea then. I kind of assumed we would want an assertion, but not sure whyI did that.

bdmendes:
I'll eventually try to propose something then (pun intended). I have some implementation doubts, it does not seem to be very simple.

[...]

PR #1162, opened 2026-08-10
Add `eventually` utility

bdmendes:
Hi! As discussed in https://github.com/scalameta/munit/issues/1161, here's a proposal for `eventually`. Some things to think about:

- I have experimented with an effect agnostic version that did `munitValueTransform(body)` and returned a `Future`. However, that wouldn't work for sync assertions before the end of a test example, and wouldn't return composable effect types for the interested downstream effect adapter libraries.
- I don't think we have a way to sleep on JS for the synchronous case. Am I missing something?
- The type implementations catch all exceptions and retry, not only assertion failures. This makes it very simple to e.g. skip transient failures on a Future and only map on the happy value. I understand this is opinionated, let me know what you think (vs. only catching munit assertion failures).
- The stack trace on failures is very long (recursive retries). Should we clean it up and filter the retry calls?

I am open to all kinds of suggestions.

### DIFF
--- a/munit/shared/src/main/scala/munit/EventuallyTransforms.scala
+++ b/munit/shared/src/main/scala/munit/EventuallyTransforms.scala
@@ -1,0 +1,12 @@
+package munit
+
+trait EventuallyTransforms {
+  this: BaseFunSuite =>
+
+  /** Evaluates the effectful body until it succeeds or max retries are exhausted. */
+  def eventually[A](
+      body: => A
+  )(implicit options: EventuallyOptions, transform: EventuallyTransform[A]): A =
+    options.eventually(body)
+
+}
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -496,6 +496,64 @@ use historical
 to keep track of how frequently flaky tests are failing and to get a better
 understanding of when and why they are failing.
 
+## Retry failing assertions
+
+Use `eventually()` to re-evaluate a body until its assertions stop failing, for
+a test that waits on something outside its control such as a background write or
+an eventually-consistent read.
+
+In order to use it, you must first build an instance of `EventuallyOptions` with
+`EventuallyOptions(retries, sleep)` (both parameters must be positive) and modify
+using `withMaxRetries`, `withSleep`, `withFilters` (or `addFilters`). You can
+also use `EventuallyOptions.disabled`, which runs the body without retrying.
+
+> On Scala.js a body that is not async cannot sleep between attempts, so its
+> retries run back to back.
+
+The filters are predicates taking a `Throwable` and returning a Boolean, to
+designate which exceptions we will retry on. By default, only failed munit
+assertions (derived from `munit.FailExceptionLike`) are retried. `addFilters`
+adds filters to the current set, while `withFilters` replaces it.
+
+`eventually()` needs an `EventuallyOptions` in implicit scope. (As is usual
+with implicits, if you declare one for the whole suite and want to override it
+in a narrower scope, you must give the inner one the same name.) You can also
+call `eventually()` directly on an `EventuallyOptions` when a single call
+needs its own settings.
+
+```scala mdoc
+import munit.EventuallyOptions
+
+import scala.concurrent.duration._
+
+class EventuallySuite extends munit.FunSuite {
+  private implicit val options: EventuallyOptions =
+    EventuallyOptions(40, 100.millis)
+
+  // a query against a table another process writes to
+  def rowCount: Int = ???
+
+  test("rows arrive") {
+    eventually(assertEquals(rowCount, 3))
+  }
+
+  test("this table takes longer to fill") {
+    EventuallyOptions(60, 1.second).eventually(assertEquals(rowCount, 3))
+  }
+
+  test("the connection also comes up late") {
+    options.addFilters(_.isInstanceOf[java.io.IOException])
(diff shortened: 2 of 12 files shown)


tgodzik:
Looks like a sensible addition to me, but maybe we should ask for more opinions? I can raise it on Scala contributors and discord to get more eyes on it. What do you think?

tgodzik:
> The stack trace on failures is very long (recursive retries). Should we clean it up and filter the retry calls?

One way hand yes, but I find tinkering with stack trace might cause more issues and is a bit magical.

bdmendes:
> I can raise it on Scala contributors and discord to get more eyes on it

Sure, let's try to find some common ground. Can you point me to the Scala Contributors thread/Discord where the discussion happens? I am happy to keep an eye on those and in this PR's thread.

tgodzik:
I've posted on the general [Scala Discord](https://discord.gg/4by8yP3Kf) and on [Scalameta one](https://discord.gg/texK9xAZe). I think ideally people would comment here.

I just wanted some additional opinions, since I mostly maintain munit in a minimal capacity 😓 (didn't work on the library to extensively)

kitbellew:
> * I have experimented with an effect agnostic version that did `munitValueTransform(body)` and returned a `Future`. However, that wouldn't work for sync assertions before the end of a test example, and wouldn't return composable effect types for the interested downstream effect adapter libraries.

i didn't understand this part.

I'm pretty sure everything we do in this library is a Future (or you wouldn't have had to add an `await`). which means, this approach is inconsistent and needs to be revised.

the user needs to assert inside the body, or worry about await themselves. you just need to avoid retrying failures due to failed assertions (this looks like skipping a particular exception type).

bdmendes:
> I'm pretty sure everything we do in this library is a Future

I am not getting your point here. munit assertions are by design "synchronous" and run wherever the user wants them to run. It is perfectly wise to do

```scala
test("do something"):
  assertEquals(1,1)
  assert(1 == 1)
```

And none of these return a `Future`. Behind the scenes, munit will try to transform the evaluated `assert(1 == 1)` through the value transforms, and here does nothing as it's Unit. It will do work when the return value is a `Future`, or something the user has provided how to convert to it.

The whole transformer mechanism is of course evident when

```scala
test("do something"):
  for
    v <- fut()
  yield assertEquals(v, 1)
```

But then again, the first example is also perfectly valid.

The goal of this proposal is for `eventually` to be an "effect transformer" that retries failed futures, as for instance https://github.com/cb372/cats-retry. This PR does await on Unit, as it takes Unit as an effect. What I was getting with

>  have experimented with an effect agnostic version that did munitValueTransform(body) and returned a Future

Is that if `eventually` is 

```scala
def eventually[A](body: Any) = withRetries(() =>munitValueTransform(body))
```

Then we lose the capacity to easily retry synchronous code as the test examples in this PR show. Also, what framework integrations consider their preferred way to convert to the Future world may necessarily be their preferred way of retrying failures. This design allows them to do it inside of their effect world before the final Future conversion runs.

> you just need to avoid retrying failures due to failed assertions

The goal of this is to retry failed assertions, to wait until something that we expect to happen eventually happens. Eg, something is listening in a Docker container port, etc. This is an alternative take of [specs2 eventually](https://etorreborre.github.io/specs2/api/SPECS2-5.9.1/org/specs2/matcher/EventuallyMatchers$.html). Of course, we can catch only assertion errors, instead of most exceptions, that's a decision choice.

Please, explain your point further.

kitbellew:
> Please, explain your point further.

i saw your `await` while reviewing on the phone and somehow interpreted as equivalent to `Await.result`.

kitbellew:
@bdmendes @tgodzik i revised the change. the thing that bothered me the most had to do with retrying any exception, so i fixed that. now by default it only retries failed assertions, but this can be overridden. it can also be overridden per assertion or per block.

munit/shared/src/main/scala/munit/EventuallyTransforms.scala
```diff
@@ -0,0 +1,19 @@
...
+   *
+   * Uses `munitEventuallyOptions` unless an `EventuallyOptions` is implicitly
+   * in scope, which lets a block override them for the calls it encloses.
+   */
+  def eventually[A](body: => A)(implicit
+      options: EventuallyOptions = munitEventuallyOptions,
+      transform: EventuallyTransform[A],
+  ): A = options.eventually(body)
+
+  def munitEventuallyOptions: EventuallyOptions = EventuallyOptions.disabled
```
bdmendes: What's the point of this default? Doesn't this render `eventually` a noop?

albertikm: correct. there's no "reasonable" default, in my opinion.

bdmendes: In that case, why try to provide one? This makes `eventually` similar to `identity`, right? I think it's reasonable to require the user to provide retry settings, either via implicit or explicitly.

kitbellew: sounds good, let's do it this way. @tgodzik please chime in (since i revised it, i guess it's my implicit "ok".)

bdmendes: Thanks. This looks okay to me.

tgodzik (approved):
LGTM from me!

PR #1168, opened 2026-08-25
fix: handle null exception message in munitAppendToFailureMessage

### LINKED ISSUE #1090: munitAppendToFailureMessage NPEs when the test exception has a null message
Hit this while adding `munitAppendToFailureMessage` to a couple of suites for extra debug context.

If a test throws an exception with a null message (`new NullPointerException()`, or a custom `Throwable` that forgot to pass one up), the failure-suffix logic itself throws an NPE. The original test exception gets swallowed and the stack trace ends up pointing at munit, not at the test.

Reproducer (scala-cli):

```scala
//> using scala 3.8.3
//> using dep "org.scalameta::munit:1.3.0"

import munit.FunSuite

class NullMessageSuite extends FunSuite:
  override def munitTestTransforms =
    super.munitTestTransforms ++ List(
      munitAppendToFailureMessage(_ => Some("extra context"))
    )

  test("trips on null exception message") {
    throw new NullPointerException() // no message
  }
```

Output (trimmed):

```
==> X NullMessageSuite.trips on null exception message
java.lang.NullPointerException: Cannot invoke "String.endsWith(String)" because "existing" is null
  at munit.TestTransforms.append$1(TestTransforms.scala:68)
  ...
```

The `append` helper inside `munitAppendToFailureMessage` looks like the source:

https://github.com/scalameta/munit/blob/591eb2c92a3e130e6f20df458ae047195e84c838/munit/shared/src/main/scala/munit/TestTransforms.scala#L58-L80

Line 74 passes `e.getMessage` straight into `append`, and `append` on line 68 calls `existing.endsWith("\n")` without checking for null. For exceptions with no message that's an NPE. The `FailExceptionLike` branch right above doesn't hit this — it goes through `updateMessage`, which operates on whatever string is already there.

Wrapping the message before `append` would fix it: `Option(e.getMessage).getOrElse(e.getClass.getName)` or similar. Happy to PR if you'd like.

arimu1:
## Summary

- Guard `munitAppendToFailureMessage` against throwables whose `getMessage` is `null` by falling back to the exception class name before calling `append`.
- Add a framework test (`suffix-null-message`) that throws a message-less `NullPointerException` with a failure suffix transform applied.

Fixes #1090

## Test plan

- [x] Reproduced pre-fix NPE: `Cannot invoke "String.endsWith(String)" because "existing" is null` when `suffix-null-message` runs without the guard
- [x] `sbt preparePR` (scalafmt, javafmt, scalafixCheckAll)
- [x] `sbt "tests2_13/testOnly munit.FrameworkSuite"` — 39 passed, 0 failed

### DIFF
--- a/munit/shared/src/main/scala/munit/TestTransforms.scala
+++ b/munit/shared/src/main/scala/munit/TestTransforms.scala
@@ -71,7 +71,8 @@ trait TestTransforms {
               Failure(Exceptions.rootCause(exception) match {
                 case fel: FailExceptionLike[_] => fel.updateMessage(append)
                 case e => new FailException(
-                    message = append(e.getMessage),
+                    message =
+                      append(Option(e.getMessage).getOrElse(e.getClass.getName)),
                     cause = e,
                     isStackTracesEnabled = false,
                     location = t.location,
--- a/tests/shared/src/main/scala/munit/TestTransformFrameworkSuite.scala
+++ b/tests/shared/src/main/scala/munit/TestTransformFrameworkSuite.scala
@@ -17,6 +17,7 @@ class TestTransformFrameworkSuite extends munit.FunSuite {
   test("suffix-success") {}
   test("suffix-fail")(fail("boom"))
   test("suffix-assertEquals")(assertEquals(0, 1))
+  test("suffix-null-message")(throw new NullPointerException())
 }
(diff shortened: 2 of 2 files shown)


tgodzik (approved):
LGTM

